PR #3, opened 2016-03-20
Add support for IoC containers

codyjs:
I've been using controllers.ts and I really like it, but I wanted to use it with an IoC container like [Inversify](inversify.io). I dug through the code and saw that container support was implemented but there was no way to provide the factory with a container.

I added a sample project to test this feature out, `sample7-container`.

When setting up the sample, be sure to use the packaged code from the repository instead of the npm verson! (This might be obvious, but it took me some time to figure out). I included setup instructions in the readme.


src/Factory.ts
```diff
@@ -7,8 +8,11 @@ import {ExpressServer} from "./server/ExpressServer";
  * @param expressApp Express application instance
  * @param requireDirs Array of directories where from controller files will be loaded
  */
-export function registerActionsInExpressApp(expressApp: any, requireDirs?: string[]): ControllerRunner {
+export function registerActionsInExpressApp(expressApp: any, requireDirs?: string[], container?: Container): ControllerRunner {
```
pleerock: I think this change is not recommended, since list of parameters can be become large very soon, so only first-priority params should be used (container cannot be called so). This can be accepted for now, but in the future I'm planning to change api of this method to make it more flexible.

src/ControllerRunner.ts
```diff
@@ -12,7 +12,7 @@ import {ResponseInterceptorInterface} from "./interceptor/ResponseInterceptorInt
 import {ResponseInterceptorMetadata} from "./metadata/ResponseInterceptorMetadata";
 import {ResultHandleOptions} from "./ResultHandleOptions";
 
-export type Container = { get(someClass: any): any };
+export interface Container { get(someClass: any): any };
```
pleerock: can you please explain motivation of this change?

sample/sample7-container/tsd.json
```diff
@@ -0,0 +1,22 @@
+{
```
pleerock: I am sorry but looks like I didnt pushed latest changes, but controllers.ts is already using typings instead of tsd (since tsd is deprecated) (1 week ago I pushed to npm latest version with typings), so you'll need to update this and use typings instead

pleerock:
Hi! Thank you very much for your contribution. Didn't you try to use [typedi](https://github.com/pleerock/typedi) instead of inversify?)

pleerock:
Also, can you please make your sample same style as others? (without its own package.json, tsd.json, tsconfig and others)? like [here](https://github.com/pleerock/controllers.ts/tree/master/sample/sample1-simple-controller)

src/Factory.ts
codyjs: I can understand your concern here. Alternatively, there could be a separate method that provides the framework with a global container. This way you would not have to alter this method's API. What do you think?

src/ControllerRunner.ts
codyjs: I changed this from a type alias so that custom containers (see [InversifyContainer](https://github.com/codyjs/controllers.ts/blob/inversify/sample/sample7-container/InversifyContainer.ts)) could implement this as an interface.

codyjs:
After looking at typedi, it looks very straightforward and easy to use. Is there currently a way to use it together with controllers.ts?

Also, I could take out the tsconfig.json from the sample but the typings and package.json contain Inversify dependency. I suppose I could implement a sample that doesn't use any 3rd-party dependencies.

codyjs:
Removed to .json files from the sample directory.

src/Factory.ts
pleerock: I would like to accept this changes for now. But in the future this and some other parts will be changed anyway since the library is in early beta, I just want to warn you because you are one of the few library users (I hope you are okay with changes)

src/ControllerRunner.ts
pleerock: got it, thanks. I used type because I expected that users will use only structural matching of the types. But having interface maybe use useful for some users too

pleerock:
> Also, I could take out the tsconfig.json from the sample but the typings and package.json contain Inversify dependency. I suppose I could implement a sample that doesn't use any 3rd-party dependencies.

I can suggest you simply to add dev dependency to the project. Samples are part of the dev, thats why I think its okay to add them there. What do you think about it?

pleerock:
> After looking at typedi, it looks very straightforward and easy to use. Is there currently a way to use it together with controllers.ts?
typedi usage with controllers.ts is pretty same way as with other containers. Something like this:

```typescript
import {Container} from "typedi/Container";
let controllerRunner = new ControllerRunner(...);
controllerRunner.container = Container;
```
 
as an example I used typedi in my microframework. Microframework has modules including for controllers.ts. Controllers.ts integration with typedi goes throw microframework [here](https://github.com/pleerock/microframework-controllers.ts/blob/master/src/ControllersTsModule.ts#L105).

codyjs:
I see. If that is the preferred way of setting the container then maybe this PR is not necessary.

src/Factory.ts
```diff
@@ -1,5 +1,17 @@
 import {ControllerRunner} from "./ControllerRunner";
 import {ExpressServer} from "./server/ExpressServer";
+import {Container} from './ControllerRunner';
```
pleerock: looks like `ControllerRunner` class should be called `Container` ?

pleerock:
If you are talking about `setContainer` method - then yes, I dont like how you added to `Factory.ts`. Optional parameter were better. 

> this PR is not necessary.

but what about sample itself?

codyjs:
I will change it back.

I realized that if a user provides a container, this means that all of the classes within it have already been loaded, so in this case the `requireDirs` parameter is unnecessary.

To fix this, there are two options: overload the method or add an options parameter, like 

`{ requireDirs?: string, contanier?: Contanier}`

Which would you prefer?

pleerock:
> I realized that if a user provides a container, this means that all of the classes within it have already been loaded, so in this case the requireDirs parameter is unnecessary.

It depends of how container is implemented. For example typedi create class instances, as they are "required", and you dont need to "setup classes to container", it will create instances and inject if needed only when its necessary - e.g. after controller.ts loaded its controllers via "requireDirs"

> Which would you prefer?

better to make options parameter, this is exactly what I wanted to do there. I already applied this practice in some other my libraries, [like here](https://github.com/pleerock/typeorm/commit/c2b73f0811b34c3099a6c57ccb7011d7cde01493#diff-c038855279ef82c6b92fad9b1725e008R13)

pleerock:
oh @codyjs please forgive me, I completely forgot about this PR. are you still interested in these changes?

pleerock:
container now can be easily specified in options into createExpressServer / useExpressServer functions

PR #6, opened 2016-05-11
add middleware and next function support, see sample8 for example

Salimlou:
this PR add the ability to specify middleware, and also Inject Next function if needed.
 there is also some whitespaces thant tslint showed.

### DIFF
--- a/sample/sample8-middlewares/UserController.ts
+++ b/sample/sample8-middlewares/UserController.ts
@@ -1,0 +1,56 @@
+import {Request} from "express";
+import {Controller} from "../../src/decorator/controllers";
+import {Get, Post, Put, Patch, Delete, Use} from "../../src/decorator/methods";
+import {Req, Next} from "../../src/decorator/params";
+
+@Controller()
+export class UserController {
+
+
+    @Use()
+    middleware( @Next() next: any) {
+      console.log("calling middleware");
+      next();
+    }
+
+    @Get("/users", { jsonResponse: true })
+    getAll() {
+        return [
+            { id: 1, name: "First user!" },
+            { id: 2, name: "Second user!" }
+        ];
+    }
+
+    @Get("/users/:id")
+    getOneMdlw( @Req() request: Request, @Next() next: any) {
+        console.log("calling next");
+        next();
+    }
+
+    @Get("/users/:id")
+    getOne( @Req() request: Request) {
+        return "User #" + request.params.id;
+    }
+
+    @Post("/users")
+    postmdlw( @Req() request: Request) {
+        let user = JSON.stringify(request.body); // probably you want to install body-parser for express
+        return "User " + user + " !saved!";
+    }
+
+    @Put("/users/:id")
+    put( @Req() request: Request) {
+        return "User #" + request.params.id + " has been putted!";
+    }
+
+    @Patch("/users/:id")
+    patch( @Req() request: Request) {
+        return "User #" + request.params.id + " has been patched!";
+    }
+
+    @Delete("/users/:id")
+    remove( @Req() request: Request) {
+        return "User #" + request.params.id + " has been removed!";
+    }
+
+}
--- a/sample/sample7-parsed-models/Photo.ts
+++ b/sample/sample7-parsed-models/Photo.ts
@@ -3,7 +3,7 @@ export class Photo {
     id: number;
 
     url: string;
-    
+
     isUrlEmpty() {
         return !this.url;
     }
--- a/sample/sample7-parsed-models/User.ts
+++ b/sample/sample7-parsed-models/User.ts
@@ -2,15 +2,15 @@ import {Photo} from "./Photo";
 import {Type, Skip} from "constructor-utils/constructor-utils";
 
 export class User {
-    
+
     id: number;
-    
+
     name: string;
 
     @Skip()
     password: string;
-    
+
     @Type(() => Photo)
     photo: Photo;
-    
+
 }
--- a/sample/sample7-parsed-models/UserController.ts
+++ b/sample/sample7-parsed-models/UserController.ts
@@ -18,5 +18,5 @@ export class UserController {
         console.log("Is photo url empty?: ", user.photo.isUrlEmpty());
         return user;
     }
-    
+
 }
--- a/sample/sample7-parsed-models/UserFilter.ts
+++ b/sample/sample7-parsed-models/UserFilter.ts
@@ -1,10 +1,10 @@
 
 export class UserFilter {
-    
+
     keyword: string;
-    
+
     hasKeyword(): boolean {
         return this.keyword && this.keyword.length > 2;
     }
-    
+
 }
--- a/sample/sample8-middlewares/app.ts
+++ b/sample/sample8-middlewares/app.ts
@@ -1,0 +1,10 @@
+import * as express from "express";
+import {registerActionsInExpressApp} from "../../src/routing-controllers";
+
+require("./UserController");
+
+let app = express(); // create express server
+registerActionsInExpressApp(app); // register controllers routes in our express application
+app.listen(3001); // run express app
+
+console.log("Express server is running on port 3001. Open http://localhost:3001/users/");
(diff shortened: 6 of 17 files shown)


sample/sample8-middlewares/UserController.ts
```diff
@@ -0,0 +1,56 @@
...
+import {Controller} from "../../src/decorator/controllers";
+import {Get, Post, Put, Patch, Delete, Use} from "../../src/decorator/methods";
+import {Req, Next} from "../../src/decorator/params";
+
+@Controller()
+export class UserController {
+
+
+    @Use()
+    middleware( @Next() next: any) {
```
pleerock: as I understood this method `middleware` will be called before each route in this specific controller, am I right?

sample/sample8-middlewares/UserController.ts
```diff
@@ -0,0 +1,56 @@
...
+    @Get("/users", { jsonResponse: true })
+    getAll() {
+        return [
+            { id: 1, name: "First user!" },
+            { id: 2, name: "Second user!" }
+        ];
+    }
+
+    @Get("/users/:id")
+    getOneMdlw( @Req() request: Request, @Next() next: any) {
```
pleerock: not sure if the `@Next()` is the best name here... Also looks like this decorator makes this library more express-specific

pleerock:
I think two main questions here to discuss:
* how do we make application-level middleware?
* how do we make route-level middleware?

sample/sample8-middlewares/UserController.ts
Salimlou: Use decorator acts like actions one(GET, POST,..) but instead of calling express[action-type](router...) while registering action, I use express.use method, so the same rules applies, in this example the middleware will be called before **all route in all controllers**, since the calculated route is "/".
the main idea here is that the middlewares are controller agnostic, the decision to call them or not depends on the provided and calculated route.

sample/sample8-middlewares/UserController.ts
Salimlou: I agree, it tied it up with express, perhaps we can specify some specific decorator in server implementations, to that they're only avalaible when a specific server framework is used.

I did a quick search for next function in koa, it seem they use it the same way... for the others I really don't know :)!

If you have suggestions?

Salimlou:
> how do we make application-level middleware?

the calculated route should be /, like in the first example.

for instance, I used with passport this way : 

```
    @Use()
    passportMdl( @Req() request: Request, @Res() response: Response, @Next() next) {
      passport.authenticate("jwt", { session: false })(request, response, next);
    }
```

> how do we make route-level middleware?

``` 
    @Use("/users/:id")
    getOneMdlw( @Req() request: Request, @Next() next: any) {
        console.log("calling next");
        next();
    }

    @Get("/users/:id")
    getOne( @Req() request: Request) {
        return "User #" + request.params.id;
    } 
```

One important thing that I forgot! Method defintion order is **very important**, there is no sorting.... for now :)

sample/sample8-middlewares/UserController.ts
Salimlou: I used Next in case we need middleware for a specific HTTP verb.

In this case we use the method decorator (Get, Post,...) and the Next function for passing to the next route...

sample/sample8-middlewares/UserController.ts
pleerock: if was afraid that this middleware will be called for `all route in all controllers` because it will lead to lot of problems. When you have user module with its own controller and lets say posts controller in the posts module and it brings *global* middleware which changes something, and you need to think why user controller actions works unexpectedly. If this middleware is for all controllers, then it should not be in specific controller.

sample/sample8-middlewares/UserController.ts
pleerock: I haven't used next function a lot, and I can't get one point... Do we really need to use @Next here? I makes sense for me if we used it in something that is running before this controller action (`getOneMdlw`), but this controller action getOneMdlw is the last in the stack, and there is nothing to call after it using next. Or am I wrong?

pleerock:
This is what I can suggest:

> application-level middleware?

```typescript
@Middleware("middlewareName")
export class SomeMiddleware implements MiddlewareInterface {
      use(request, response, next) { 
            // do something
            next();
      }
}
```
signatures in the libraries:
```typescript

export function Middleware(name?: string, options?: { priority?: number, routes?: string[] });

export interface MiddlewareInterface {
      use(request, response, next): void;
}
```

> how do we make route-level middleware?

```typescript
@Use("middlewareName") // can be applied on both controller class and controller method
@Get("/users/:id")
getOne( @Req() request: Request) {
    return "User #" + request.params.id;
} 
```

sample/sample8-middlewares/UserController.ts
Salimlou: it's the getOne method that will be called., the getOneMdlw method acts like middleware in this case.

```
   @Use("/users/:id") //  or @Get("/users/:id")
    getOneMdlw( @Req() request: Request, @Next() next: any) {
        console.log("calling next");
        next();
    }

    @Get("/users/:id")
    getOne( @Req() request: Request) {
        return "User #" + request.params.id;
    } 
```

Salimlou:
Nice, i like it, but I have a question : 

1. Declaring a Middleware class, make it de facto application wide. with routes options we can specify routes to which this middleware will be applied. Do you really need to use @Use Decorator to have route-level middleware in this case?


wa can introduce a scope parameter, that the registrator will handle this way: 

- if the scope is application, the middleware is registred before all the controllers
- if the scope is route, it the will be registred before the route or controller that have a **@Use("mdlName")** decorator

```
@Middleware("appWideMiddleware", {scope : SCOPE.APPLICATION})
export class AppWideMiddleware implements MiddlewareInterface {
      use(request, response, next) { 
            // do something
            next();
      }
}

@Middleware("routeMiddleware", {scope : SCOPE.ROUTE})
export class RouteMiddleware implements MiddlewareInterface {
      use(request, response, next) { 
            // do something
            next();
      }
}
```
in this case we don't need options.routes .

So, if you want I can start implementing your design, and see How it goes! I'm more of a code'n'see guy :)!

Salimlou:
Another question, this decorator will be express specific??

pleerock:
> with routes options we can specify routes to which this middleware will be applied. Do you really need to use @Use Decorator to have route-level middleware in this case?

Actually `@Use` should be used instead of `options.routes` in a middleware decorator in most of cases. `options.routes` since its global should be used in rear cases, only in some situation =)

> scopes
options.routes is list of routes for which this middleware will work. I dont see how scopes can handle this.

> Another question, this decorator will be express specific??

no. I think koa has same concepts and it can be applied on it too. 

>  I can start implementing your design

I think better if I do it plus I will make some refactorings that I wanted to do. But if you want to contribute would be great if you implement koa or other framework support, or take care of any other feature.

Salimlou:
Okay Sir :+1:  

I never worked with koa or any other server, so I don't have enough perspective like with express.

I've been working with this lib for more than 6 months now, and the only thing I needed the most is the middleware stuff for passport.js

If you need something particular let me know. 
And keep up the good work, I learned a lot for your code :)!!

pleerock:
thank you, I think I'll implement middlewares on this weekends. By the way, do you use relational databases (mysql / postgre / etc.)  in your projects?

Salimlou:
Cool. 

For relational databases, I haven't used them for over a year now, since my former Job. At that time it was mainly Java/J2EE and Oracle. 

Now I'm all NoSQL, mainly Mongo, with mongoose...., But perhaps some potential future projects will need a RDB?? 

why? Do you have a request? something I can help with?

on the other side, should i close this pull request??

pleerock:
no, just wanted to point out my last project - [typeorm](https://github.com/pleerock/typeorm). I also was working on [mongodb odm](https://github.com/pleerock/typeodm) but for now freeze the work, planning to return back to it later after finish with orm. Try them (use samples)

Salimlou:
I will :)!!

pleerock:
closing it since it was already implemented

PR #9, opened 2016-05-31
"Mount" a directory under a specific path

alexproca:
Added option to "mount" controllers from a directory under a specific path


pleerock:
thank you

PR #13, opened 2016-07-13
0.6.0 release

pleerock:


pleerock:
added koa support special for @amcdnl . only noticable thing that is not working now is file uploading. need to find a good uploading middleware and implement that feature too

PR #22, opened 2016-08-17
Fixes Error Handler documentation

JD-Robbs:
This fixes an issue where the documentation and code have got out-of-sync.

As far as I know, the Error Handler Middleware now does not need to be imported anymore since it should be loaded by use/createExpressServer's `middlewareDirs`.  That being said, maybe not everyone uses that option and may still need to manually include the files.

I am referring to the following:

    -2. Load created error handler before app bootstrap:		
    -		
    -    ```typescript		
    -    import "reflect-metadata";		
    -    import {createExpressServer} from "routing-controllers";		
    -    import "./UserController";		
    -    import "./CustomErrorHandler"; // here we load it		
    -    createExpressServer().listen(3000);		
    -    ```		
    -

Long story short: maybe the above shouldn't be removed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -722,15 +722,15 @@ You can make global middleware to run after controller action by using `@Middlew
 ### Error handlers
 
 Error handlers are specific only to express.
-Error handlers works pretty much the same as middlewares, but instead of `@Middleware` decorator `@ErrorHandler` is being used:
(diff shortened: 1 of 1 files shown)


pleerock:
thank you

Issue #30, opened 2016-09-17
Session Method Parameter Decorator


19majkel94:
Is the `@Session` decorator avaible?
I attach to `req.session` an user object after sucessfull authentication and use it in controler method to check if user is an owner of demanded object from database.
It would be nice to have :

    @Get("/posts")
    getAll(@Session("user") user: User) {
       return postsRepository.getAllByUser(user);
    }

And don't have to parse request object by myself :wink:

pleerock:
No, there is no session decorator yet. I can implement it, but not now, pretty busy with other repositories. Or if someone else can implement it, it would be great, implementation should be easy. For now please use req.session :(

19majkel94:
If it's all about add Session decorator in `src/decorator/params.ts`, add Session type in `src/metadata/types/ParamTypes.ts` and implement behaviour in `src/ParamHandler.ts` - I think I could do that but first I have to build an app using your stack and then test if the changes will work :smile:

pleerock:
Great! 👍

pleerock:
thanks @19majkel94 . Hope someone take care of it in koa too

PR #31, opened 2016-09-18
Added @Session decorator

19majkel94:
Implemented issue #30.
Works for injecting whole session object - `@Session() session: Express.Session` or with single object from session - `@Session("objectName") object: ObjectType`.

### DIFF
--- a/test/functional/action-params.spec.ts
+++ b/test/functional/action-params.spec.ts
@@ -102,6 +102,11 @@ describe("action parameters", () => {
             }))
             addToSession(@Session() session: Express.Session) {
                 (session as any).testElement = "@Session test";
+                (session as any).fakeObject = {
+                    name: "fake",
+                    fake: true,
+                    value: 666
+                };
                 return `<html><body>@Session</body></html>`;
             }
 
@@ -307,7 +312,6 @@ describe("action parameters", () => {
             expect(response).to.have.header("content-type", "text/html; charset=utf-8");
             expect(response.body).to.be.equal("<html><body>@Session</body></html>");
             assertRequest([3001], "get", "session", response => {
-                paramUserId.should.be.equal(1);
                 expect(response).to.be.status(200);
                 expect(response).to.have.header("content-type", "text/html; charset=utf-8");
                 expect(response.body).to.be.equal("<html><body>@Session test</body></html>");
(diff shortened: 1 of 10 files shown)


19majkel94:
I wonder if we could use the param name from reflect to not duplicate param name in quotes - `@SessionObject() user: User` could look for `req.session.user`?

19majkel94:
@pleerock Should I add `express-session` to package.json and typings.json or remove types from new samples?

pleerock:
> I wonder if we could use the param name from reflect to not duplicate param name in quotes - @SessionObject() user: User could look for req.session.user?

I think you can't

> Should I add express-session to package.json and typings.json or remove types from new samples?

you can add only to dev dependencies, since this library isn't tied to express, it can be used with koa too, and with koa users don't need express in their dependencies

pleerock:
Some things left to do:
* add note in readme that users need to install `express-session` to use `@Session` decorator
* provide good tests
* support koa too, because koa users may use `@Session` decorator too, and what they will get if they use it?

19majkel94:
I don't know koa, so I added throwing error like on `@UploadedFile`.
Readme updated about not support koa yet + `express-session` dependency.

pleerock:
please provide test coverage

19majkel94:
Added tests :smile: 
Only the error throwing for koa is not tested.

test/functional/action-params.spec.ts
```diff
@@ -91,6 +96,29 @@ describe("action parameters", () => {
                 return `<html><body>${firstId},${secondId}</body></html>`;
             }
 
+            @Post("/session/")
+            @UseBefore(session({
+                secret: "19majkel94_helps_pleerock",
+            }))
+            addToSession(@Session() session: Express.Session) {
+                (session as any).testElement = "@Session test";
+                (session as any).fakeObject = {
```
19majkel94: `fakeObject`? I added it just to be sure that the `@Session("testElement")` return the correct object, not the whole session container - the coverage report shows that the line 154 line `return request.session[param.name];` is not covered which is wrong because it's tested and works ;)

19majkel94:
>> I wonder if we could use the param name from reflect to not duplicate param name in quotes - `@SessionObject() user: User` could look for `req.session.user`?

> I think you can't

Angular's Dependency Injection can read this info using regex and `toString()`:
http://stackoverflow.com/a/12108723/5472617
So it's possible but complicated - it's better to wait for enhancement of Typescript Reflect API.

pleerock:
The problem is that when code is minimized it changes all property names too, and it can be a problem

PR #34, opened 2016-09-25
Update README.md

ragnarol:
Just a small comment clarifying the JSONController.

It took me a while to figure out why the bode of my requests with "application/json" weren't parsed into the @body() parameters. I went through the documentation and saw the JSON controller but as it only talked about returning content I ignored it.

It wasn't until I had to return JSON that I changed the controller and voiala, it magically started understanding JSON content :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -257,11 +257,11 @@ export class UserController {
 }
 ```
 
-#### Output JSON instead of regular text content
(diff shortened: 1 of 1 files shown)


pleerock:
thank you

ragnarol:
No worries, thank you for this awesome library! :)

Issue #23, opened 2016-08-19
custom error middleware always execute after default


Diluka:
I have custom error middleware to handle errors and decide different way to response. But neither Infinity nor -Infinity I set to the `priority` option, the default error middleware send the response with status 500 before my error middleware. My error middleware seems to be useless.

how can I solve this?

Diluka:
Seems it's the same pleerock/routing-controllers#17

Diluka:
```
ExpressDriver.prototype.handleError = function (error, action, options) {
    options.next(error);
};
```
@peterjp80
This is the temporary way to disable the default error handler inside. This is invoked immediately atfer action raise an error and send the respone, so the next error middlewares can do nothing but just logging.

19majkel94:
@Diluka is that still not working for you? Can this issue be closed, the PR #36 has solved your problem?

PR #36, opened 2016-10-01
Fixed always default error handling

### LINKED ISSUE #17: ErrorHandler decorator
The documentation mentions using the `@ErrorHandler()` decorator, but that decorator doesn't exist.

19majkel94:
Fixed #17 and #23.

Repaired executing default error handling even when it's turned off, which made impossible defining own error handler.

Exported `HttpError` to simplify using default error handler by throwing error in controller:
```typescript
@Post("/login")
async loginUser(@Session() session: Express.Session, @Body({ required: true }) bodyUser: BodyUser) {
    if (typeof bodyUser.email !== "string" || typeof bodyUser.password !== "string") {
        throw new HttpError(400, "The email and password fields must be string!");
    }

    const dbUser = await this.userRepository.getByEmail(bodyUser.email);
    if (dbUser && await dbUser.verifyPasswordAsync(bodyUser.password)) {
        session.user = dbUser;
        return "Logged succesfully!";
    } else {
        throw new HttpError(400, "Incorrect email or password!");
    }
}
```


src/index.ts
```diff
@@ -121,6 +121,7 @@ export * from "./decorator/controllers";
 export * from "./decorator/decorators";
 export * from "./decorator/methods";
 export * from "./decorator/params";
+export * from "./error/http/HttpError";
```
pleerock: I think we can add export for all other errors too, to make use of them convenient too

pleerock:
Can you please add tests for the issues this PR resolves?

src/index.ts
19majkel94: You are right! I've done it in 6d16e78.
Now you can import them also this way:
`import { HttpError } from "routing-controllers/error";`

19majkel94:
> Can you please add tests for the issues this PR resolves?

It's hard to easy build a test in your `express-error-handling.spec.ts` - I need to disable default error handler in `createExpressServer` options and register new `MiddlewareGlobalAfter`, which might break your other tests. I would have to create a new test file for this case, is that ok for you?

pleerock:
try the best possible solution

pleerock:
released in 0.6.6

PR #38, opened 2016-10-08
Define with exactly type for express

TonyPythoneer:
Hi @pleerock 

I suggest createExpressServer should return `express.Express` type and ExpressDriver.express sets  `express.Express` type.
I don't want when I call the **createExpressServer** function, it will return an `any` value. In TypeScript,  It's not convenience to me if I need the value show its properties.

So, I improve the type definition accuracy and avoid to use `any` type on argument and return value.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -1,3 +1,4 @@
+import * as express from "express";
 import {MetadataArgsStorage} from "./metadata-builder/MetadataArgsStorage";
 import {importClassesFromDirectories} from "./util/DirectoryExportedClassesLoader";
 import {RoutingControllerExecutor} from "./RoutingControllerExecutor";
@@ -14,7 +15,7 @@ import {RoutingControllersOptions} from "./RoutingControllersOptions";
 /**
  * Registers all loaded actions in your express application.
  */
-export function useExpressServer<T>(expressApp: T, options?: RoutingControllersOptions): T {
+export function useExpressServer(expressApp: express.Express, options?: RoutingControllersOptions): express.Express {
     createExecutor(new ExpressDriver(expressApp), options || {});
     return expressApp;
 }
@@ -22,7 +23,7 @@ export function useExpressServer<T>(expressApp: T, options?: RoutingControllersO
 /**
  * Registers all loaded actions in your express application.
  */
-export function createExpressServer(options?: RoutingControllersOptions): any {
+export function createExpressServer(options?: RoutingControllersOptions): express.Express {
     const driver = new ExpressDriver();
     createExecutor(driver, options || {});
     return driver.express;
--- a/src/driver/ExpressDriver.ts
+++ b/src/driver/ExpressDriver.ts
@@ -1,3 +1,4 @@
+import * as express from "express";
 import {HttpError} from "../error/http/HttpError";
 import {UseMetadata} from "../metadata/UseMetadata";
 import {MiddlewareMetadata} from "../metadata/MiddlewareMetadata";
@@ -23,11 +24,11 @@ export class ExpressDriver extends BaseDriver implements Driver {
     // Constructor
     // -------------------------------------------------------------------------
 
-    constructor(public express?: any) {
+    constructor(public express?: express.Express) {
         super();
         if (require) {
             if (!express) {
-                try {
+                try {                  
                     this.express = require("express")();
                 } catch (e) {
                     throw new Error("express package was not found installed. Try to install it: npm install express --save");
@@ -77,7 +78,7 @@ export class ExpressDriver extends BaseDriver implements Driver {
                    interceptors: InterceptorMetadata[],
                    executeCallback: (options: ActionCallbackOptions) => any): void {
         const expressAction = action.type.toLowerCase();
-        if (!this.express[expressAction])
+        if (!(this.express as any)[expressAction])
             throw new BadHttpActionError(action.type);
 
         const useInterceptors = action.controllerMetadata.useInterceptors.concat(action.useInterceptors);
@@ -133,7 +134,7 @@ export class ExpressDriver extends BaseDriver implements Driver {
         const expressParams: any[] = [fullRoute, ...preMiddlewareFunctions, ...defaultMiddlewares, routeHandler, ...postMiddlewareFunctions];
 
         // finally register action
-        this.express[expressAction](...expressParams);
+        (this.express as any)[expressAction](...expressParams);
     }
 
     registerRoutes() {


pleerock:
Hello @TonyPythoneer thank you for your PR. I agree with your point that 

> I don't want when I call the createExpressServer function, it will return an any value. In TypeScript, It's not convenience to me if I need the value show its properties.

But I didn't include express typings because they will require user to install another dependency - `express` from `@typings` repository. If I accept this PR, probably it should include dependency to `@types/express`. Is it a good practice to have a dependency on `@types`? Adding @19majkel94

19majkel94:
> But I didn't include express typings because they will require user to install another dependency - express from @typings repository. If I accept this PR, probably it should include dependency to @types/express. Is it a good practice to have a dependency on @types? 

It's hard to say. On the one hand, `ExpressDriver` or `useExpressServer()` are tightly coupled to `express` so it should use correct `express` types. On the other hand, it's also a framework for `koa`, so having hard `express` dependency for `koa` users means that they download and install unnecessary npm packages. But if we think about unnecessary downloads, you also should distribute `*d.ts` files through `@types` repository, not include them in package.

So if we're able to have dependency in `package.json` only on`@types/express`, not the whole `express` from npm, I think it's ok :wink:

src/index.ts
```diff
@@ -14,15 +15,15 @@ import {RoutingControllersOptions} from "./RoutingControllersOptions";
 /**
  * Registers all loaded actions in your express application.
  */
-export function useExpressServer<T>(expressApp: T, options?: RoutingControllersOptions): T {
+export function useExpressServer(expressApp: express.Express, options?: RoutingControllersOptions): express.Express {
```
TonyPythoneer: According to [this section](https://github.com/pleerock/routing-controllers#using-exist-server-instead-of-creating-a-new-one), it seems useExpressServer takes a `expressExpress` variable because `let app = express();` it will get `express.Express` value. So, I think it's `express.Express`, not `express.Application`.

pleerock:
oh I didn't think about koa. having both `@types/express` and `@types/koa` isn't a good idea. To force people to install express or koa typings separately also maybe not so convenient, also if they don't install it they'll have errors. Third option is just left it as it is. Need to choose one of them

src/index.ts
19majkel94: I know. Do you know what is inheritance in OOP? The `express.Express` extends the `express.Application`:

```typescript
interface Express extends Application {
        /**
            * Framework version.
            */
        version: string;

        /**
            * Expose mime.
            */
        mime: string;

        (): Application;

        /**
        * Create an express application.
        */
        createApplication(): Application;

        createServer(): Application;

        application: any;

        request: Request;

        response: Response;
    }
```
So if we are demanding `express.Express`, user won't be able to pass his own `express.Application`. But thanks to inheritance, if we demand `express.Application` as parameter, you can easily pass there `express.Express` from `express()` :wink:

It looks like that @pleerock didn't use `express.Express` specific methods/properties, so it's not necessary to use `express.Express`.

TonyPythoneer: I see. You are right.

TonyPythoneer:
I know it will cause this project needs to add `express` in dependencies.

I have another idea. I recommend it should discard `createExpressServer` and only offer `useExpressServer` to register `routing-controller` feature. 

In my opinion:
1. They have passed `createExecutor(...);` to get `routing-controller` feature. These functions are not significantly different.
2. `useExpressServer` is better than `createExpressServer`. It has generic type when it passes a `express.Express` or `express.Application`, it will keep it type and return the application. Relatively, `createExpressServer` only returns `any` value and not good to use.

Demo:
```
// use `useExpressServer`
// When it pass a parameter and return a value, these values are the same type.
const expressApp = express();
const app = useExpressServer(expressApp);

// use createExpressServer
// The `app` is `any` type. It's not good to use and be hinted by Typescript
const app = createExpressServer()  
```

@pleerock , how do you think? :smile:

19majkel94:
First of all, `useExpressServer` shouldn't return anything. The `expressApp` parameter is simply returned which doesn't make sense because the parameter is passed by reference, which means we can use it normally:
```typescript
const expressApp = express();
useExpressServer(expressApp);
expressApp.use(...)
```
And the generic type behave just as `any` type:
```typescript
const app = useExpressServer("blahblah");
```
But using `createExpressServer()` we can simply assert type:
```typescript
const app: express.Application = createExpressServer();
```
And we are sure that typechecking is working well :wink:

TonyPythoneer:
@19majkel94  Okay, let me clarify this PR purpose.

I hope when I use `createExpressServer`, it won't return `any` value and needn't to manually assert type by developer.

Do you have any idea to slove this? :smile:

pleerock:
I like `createExpressServer` and I mostly use it rather then `useExpressServer`. I created second one for people who want to preconfigure express. I understood your purpose and there is a point to use express typings instead of any type. I see two ways:

1. Force user to install `@types/express` alongside with `express`. The only change here is in readme - we add `npm i express @types/express` instead of `npm i express`. Of course this is a bit more complicated and annoying then simply one dependency, but overall I don't think its a problem. 

2. we just left it as it is right now. Yes, problem is not solved, and we still have to manually type cast what createExpressServer returns if we want to use it. Benefit - easier installation without extra dependency.

I guess @TonyPythoneer would vote for first option, @19majkel94 what do you think about it? Is first option acceptable?

pleerock:
@19majkel94 any thoughts?

19majkel94:
The first option is ok for me, but koa users might be forced to install express typings. If they try to compile their TS solution, they will got error about tsc not being able to found module express (typings).

Maybe it's time to split the drivers from framework? User could install the express version or koa version and doesn't bother about express dependency or typings if he just use `createServer` and don't configure express outside of `routing-controllers` 😉 It could be all-in-one packages like `npm install express-routing-controllers`.

pleerock:
ohh I was thinking about it, but it will have a high maintenance cost, plus I don't have time right now to refactor all this...

19majkel94:
I think it could be just a gulp bundle script which copy or not the driver depending on version and add a express dependency in package.json? I don't know how big and deep dependencies you have in core.

So mark this splitting/refactoring as the milestone to don't forget about it. Until then I think we can't use the express typings here.

pleerock:
yeah agree forcing express users to install koa typings and force koa users to install express typings is a bad idea. better to keep things as it is right now. Maybe in the future I'll refactor this library into two separate libraries. @TonyPythoneer for now either useExpressServer either use type casting with createExpressServer, its a best way to go for now

TonyPythoneer:
In conclusion, does it keep the same?

pleerock:
yes, I hope you understood all the problems we face if we change current implementation. Solution will be to refactor library into two separate libraries - for koa and express, but this maybe for the future

TonyPythoneer:
okay, I will use `useExpressServer` to inject `routing-controllers` into `express application`. `createExpressServer` doesn't solve my problem.

19majkel94:
We definitely need to split to two modules. Fore example, the [MiddlewareInterface](https://github.com/pleerock/routing-controllers/blob/master/src/middleware/MiddlewareInterface.ts) shows that `koa` and `express` are different and the result is that you have the interface that doesn't force you to implement anything 😜 

The temporary solution which will help in moving to splited packages is that you will have subdirectories with exported parts related to specific framework. And we can then just do:
```typescript
import { MiddlewareInterface } from "routing-controllers/express";
```
and we will get the specific express interface:
```typescript
export interface MiddlewareInterface {
    /**
     * Called before controller action is being executed.
     */
    use(request: express.Request, response: express.Response, next?: express.NextFunction): any;
}
```

pleerock:
right, I have same issue. I simply removed `implements MiddlewareInterface` from my classes after all 😞

PR #39, opened 2016-10-08
Boost maintainability

TonyPythoneer:
Hi @pleerock 

There it can use `spread` feature with options when properties of options are undefined, it can set a default value.

So, I improved the code, it will look clear.


pleerock:
Hello @TonyPythoneer and thank you for your PR. I don't think this PR brings improvements into exist code. Spread operator is good and sometimes can be really useful, but I don't think in this case. The result code is much more hardly to understand, and this is a big problem, especially in opensource where everyone is reading other's code

TonyPythoneer:
I understand. Close.

PR #48, opened 2016-11-16
missing "typings" in package.json file

aliai:
this is required for typescript to find and resolve the module and its typings :)

https://www.typescriptlang.org/docs/handbook/declaration-files/publishing.html

### DIFF
--- a/package.json
+++ b/package.json
@@ -68,5 +68,6 @@
   },
   "scripts": {
     "test": "gulp tests"
-  }
+  },
+  "typings": "index.d.ts"
 }


pleerock:
thank you for this PR. But isnt it "index.d.ts" by default when you dont specify a "main" at all?

aliai:
Well, the `main` is present, the `typings` was missing. I suppose it used to work without `typings`, but I am not entirely sure. They moved away from `typings` package management from TypeScript 2.0 if I am not mistaken, and introduced this gem as an alternative. We are using TypeScript 2.0.10 now and it complains that it cannot find the module, but after manually adding this line to the `package.json` it worked flawlessly :)

not sure why Travis is failing :(

Thanks for the great work btw!

pleerock:
sorry my bad looks like I completely forgot about this PR, just saw it still hangs here. Apologies again!

aliai:
thanks mate :)

PR #60, opened 2017-01-12
Koa Render Closes #59

### LINKED ISSUE #59: Koa render
Currently in the KoaDriver in the handleSuccess method this.koa.render is called but is undefinded.

I can't find anywhere where koa does rendering, it seems this has been copied from the ExpressDriver because Express handles its rendering, koa uses rendering middlewares that then create the render method on the context.

I am using koa-views as my renderer but getting the following error:

```
{"name":"TypeError","message":"this.koa.render is not a function","stack":"TypeError: this.koa.render is not a function\n    at KoaDriver.handleSuccess 
```

I have had a quick play around, and found that changing the call from this.koa.render to options.context.render does cause something to try and load, but timesout without any errors. I haven't looked completely through the routing-controller source to completely understand whats going on.

If anyone could give me some guidance before I dig in I would be more than happy to get this working!

joshualawson:
# Koa Render

Closes #59

## Reason

The KoaDriver render was not completed, it was using the Express way of rendering html templates using Template engines that were loaded into express.

I have decided that I prefer Koa over express so have played around with the KoaDriver to try get the render to work for Koa.

## Dependencies

I have decided to use [Koa-views](https://github.com/queckezz/koa-views) as the rendering middleware, as it allows for loading of different rendering engines easily. In the tests I have edited the Koa before section to load up the rendering middleware and load the handlebars engine by default.

## Fix

The fix I have applied is not very neat but works for now, it creates a middleware in the handleSuccess method of the KoaDriver as I needed access to Koa's context which Koa-views extends and adds the render method to. 

I tried to access the Koa context via options.context but it seem this was using routing-controllers own implementation of the context and was missing the render method. I would love suggestions to remove the middleware within the handleSuccess method and a suggestion on how I could access the Koa context correctly within the KoaDriver.

Would love your feedback @pleerock 

I hope this helps in any way!

### DIFF
--- a/README.md
+++ b/README.md
@@ -584,7 +584,9 @@ getOne() {
 To use rendering ability make sure to configure express properly.
 [Here](https://github.com/pleerock/routing-controllers/blob/0.6.0-release/test/functional/render-decorator.spec.ts)
 is a test where you can take a look how to do it.
-This feature is not supported by koa driver yet.
+
+To use rendering ability with Koa you will need to use a rendering 3rd party such as [koa-views](https://github.com/queckezz/koa-views/), 
+koa-views is the only render middleware that has been tested.
 
 ## Using middlewares
 
--- a/package.json
+++ b/package.json
@@ -53,6 +53,7 @@
     "koa": "^2.0.0",
     "koa-bodyparser": "^3.1.0",
     "koa-router": "^7.0.1",
+    "koa-views": "^5.2.0",
     "mocha": "^3.0.0",
     "multer": "^1.1.0",
     "mustache-express": "^1.2.2",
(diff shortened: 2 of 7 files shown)


pleerock:
Thank you for this clean PR. Please make sure to add sample how to use render in koa. Also probably we need to add docs [here](https://github.com/pleerock/routing-controllers#render-templates) about need of installation additional package if render is used with koa.

joshualawson:
More than happy to add in a sample and add to the docs, just wanted your input on accessing the context, as I'm not too happy about using koa.use within the handleSuccess function. Or is that my only option?

pleerock:
I don't find anything wrong with this approach, at least right now...

pleerock:
I think this can be merged, thank you for this clean job

PR #62, opened 2017-01-15
fix(useClassTransformer): fixing docs and tests

snirs90:
Docs and tests had a typo in the usage of `useClassTransformer`, was written `useConstructorUtils`.


pleerock:
Thank you!

PR #63, opened 2017-01-15
Exception throwing error fixed

emiyake:
When throwing an Exception from lower layers of the app, the correct exception wouldn't pop in the global `ErrorMiddleware`. 
Instead, the error`TypeError: Cannot read property 'handle' of undefined` would raise. This was caused by `ExpressDriver` when trying to handle the error. `this` reference was not defined properly in `ExpressDriver.registerMiddleware`. 

Using an anonymous function for the inner middleware, so that `this` is a reference to `ExpressDriver` and not a reference to the inner middleware function.

### DIFF
--- a/src/ActionCallbackOptions.ts
+++ b/src/ActionCallbackOptions.ts
@@ -8,6 +8,6 @@ export interface ActionCallbackOptions {
     /**
      * Interceptor functions to be applied for response result.
      */
-    useInterceptorFunctions: Function[];
-    
+    useInterceptorFunctions?: Function[];
+
 }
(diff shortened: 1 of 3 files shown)


pleerock:
Please add test as well, because next time if someone (including me) see `function` here he will think "why function is used here? I should change it to awesome es6 arrow syntax!" - and he will do it. To prevent this bug again we need test for this very specific case.

emiyake:
I added a test file in order to validate the custom errors, but it seems there is an issue with Travis build (not related to this commit).

PR #72, opened 2017-02-01
Fix required error handling

### LINKED ISSUE #71: ParameterRequiredError is not working
when the required key is `true` on a `@BodyParam` the error is sent as clear text (ie: not put in double quotes) and the http code is not set correctly so it becomes 500 even if that's not a server error

iH8c0ff33:
Maybe you could `throw new` instead of `return Promise.reject(new` but I'm not sure whether the exception will be handled or remain uncaught, so I decided to fix it this way.
Fixes #71 

PS: `throw new` is the only correct way

### DIFF
--- a/src/ParamHandler.ts
+++ b/src/ParamHandler.ts
@@ -1,4 +1,6 @@
 import {ParameterParseJsonError} from "./error/ParameterParseJsonError";
+import {ParameterRequiredError} from "./error/ParameterRequiredError";
+import {BodyRequiredError} from "./error/BodyRequiredError";
 import {plainToClass} from "class-transformer";
 import {ParamTypes} from "./metadata/types/ParamTypes";
 import {ParamMetadata} from "./metadata/ParamMetadata";
@@ -45,10 +47,10 @@ export class ParamHandler {
         if (param.isRequired) {
             // todo: make better error messages here
             if (param.name && isValueEmpty) {
-                return Promise.reject("Parameter " + param.name + " is required for request on " + request.method + " " + request.url);
+                return Promise.reject(new ParameterRequiredError(request.url, request.method, param.name));
 
             } else if (!param.name && (isValueEmpty || isValueEmptyObject)) {
-                return Promise.reject("Request body is required for request on " + request.method + " " + request.url);
+                return Promise.reject(new BodyRequiredError(request.url, request.method));
             }
         }
 
--- a/src/ParamHandler.ts
+++ b/src/ParamHandler.ts
@@ -47,10 +47,10 @@ export class ParamHandler {
         if (param.isRequired) {
             // todo: make better error messages here
             if (param.name && isValueEmpty) {
-                throw new ParameterRequiredError(request.url, request.method, param.name);
+                return Promise.reject(new ParameterRequiredError(request.url, request.method, param.name));
 
             } else if (!param.name && (isValueEmpty || isValueEmptyObject)) {
-                throw new BodyRequiredError(request.url, request.method);
+                return Promise.reject(new BodyRequiredError(request.url, request.method));
             }
         }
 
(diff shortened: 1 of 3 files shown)


iH8c0ff33:
Ok, that seems to break tests since they are expecting required variable not present to return 500 status instead of 400

19majkel94:
`routing-controllers` use `class-transform` to transform request params, so the handle function is asynchronous - returns a Promise. Also the param value might be a Promise:
```typescript
const promiseValue = value instanceof Promise ? value : Promise.resolve(value);
        return promiseValue.then((value: any) => {
```
But I haven't met in express that situation, that promise object is attached to request object and you have to wait for it (?!). @pleerock could you explain that part of code, do you use Promise in request in your express apps?

So if you have:
```typescript
function returningPromise(params: any): Promise<any> {
     // blah blah
    throw new Error("Throwing error");
}
```
The consumer of the function do like this:
```typescript
returningPromise().then(
    // blah blah
).catch(error => {
    // here catching promise rejection, not throwing
```
So the error is throwed synchronously and you don't catch it.

However you can throw Error in promise callback:
```typescript
function returningPromise(params: any): Promise<any> {
    return new Promise((resolve, reject) => {
        // blah blah
        throw new Error("Throwing error");
    });
}
```
Because it is wrapped in promise function, it will be transformed to Promise.reject 😉

src/ParamHandler.ts
```diff
@@ -45,10 +47,10 @@ export class ParamHandler {
         if (param.isRequired) {
             // todo: make better error messages here
             if (param.name && isValueEmpty) {
-                return Promise.reject("Parameter " + param.name + " is required for request on " + request.method + " " + request.url);
+                return Promise.reject(new ParameterRequiredError(request.url, request.method, param.name));
 
             } else if (!param.name && (isValueEmpty || isValueEmptyObject)) {
-                return Promise.reject("Request body is required for request on " + request.method + " " + request.url);
+                return Promise.reject(new BodyRequiredError(request.url, request.method));
```
19majkel94: It's weird that the `BodyRequiredError` exist in code but hasn't been used anywhere else.
Maybe @pleerock started refactoring and created it but have forgotten to change that lines 😉

iH8c0ff33:
That should have that error fixed

src/ParamHandler.ts
```diff
@@ -47,10 +47,10 @@ export class ParamHandler {
         if (param.isRequired) {
             // todo: make better error messages here
             if (param.name && isValueEmpty) {
-                return Promise.reject(new ParameterRequiredError(request.url, request.method, param.name));
+                throw new ParameterRequiredError(request.url, request.method, param.name);
```
19majkel94: You have misunderstood me... I just explained why you have to use `Promise.reject` because throwed errors will not be handled properly. See [here](https://github.com/iH8c0ff33/routing-controllers/blob/master/src/RoutingControllerExecutor.ts#L81) that there is `.catch()` chaining on promise, to sync try-catch block.

iH8c0ff33: oh... sorry 😅🙈

iH8c0ff33: is it normal that tests are failing using promise.reject?

iH8c0ff33:
ok, I found the problem @19majkel94 , or at least I think I found it, when using koa driver that line fails:
```javascript
response.status(error.httpCode);
```
because `response.status is not a function`

src/driver/KoaDriver.ts
```diff
@@ -271,7 +271,7 @@ export class KoaDriver extends BaseDriver implements Driver {
             // note that we can't use error instanceof HttpError properly anymore because of new typescript emit process
             if (error.httpCode) {
                 options.context.status = error.httpCode;
-                response.status(error.httpCode);
+                // response.status(error.httpCode);
```
19majkel94: It's really BAD when you comment part of code just to pass the test 😞 
> response.status is not a function
So change to `response.status = error.httpCode`!

iH8c0ff33: that seems to be already set somewhere else

pleerock:
Thanks @iH8c0ff33 for this PR. Thank you @19majkel94 for reviewing this PR. Can we merge it now?

PR #74, opened 2017-02-05
add simple koa session/state support

19majkel94:
@mingyang91, could you add tests for the koa features? We could merge this changes then 😉


mingyang91:
Okay, please wait a few hours

mingyang91:
Ok, please review

mingyang91:
I don't know why i got `codecov/patch — 83.33% of diff hit (target 91.32%)`.
I check Details, the `COVERAGE REACH` is all green.
😖

19majkel94:
> I don't know why i got codecov/patch — 83.33% of diff hit (target 91.32%).

Your changes has only 83.33% coverage:
https://codecov.io/gh/pleerock/routing-controllers/compare/b9e39b4eaac5481e439594bcea0f65154318a272...28c685c01f20cdd90be38f52389f0779795d8542/diff

You need to write test e.g. for throwing Error while using `@State` with Express.

test/functional/action-params.spec.ts
```diff
@@ -692,6 +696,10 @@ describe("action parameters", () => {
         assertRequest([3001/*, 3002*/], "post", "files-with-limit", undefined, invalidRequestOptions, response => {
             expect(response).to.be.status(500);
         });
+
+        assertRequest([3002], "post", "files-with-limit", undefined, validRequestOptions, response => {
+            expect(response).to.be.status(500);
```
19majkel94: why 500 when validRequestOptions?

src/driver/KoaDriver.ts
```diff
@@ -161,8 +161,20 @@ export class KoaDriver extends BaseDriver implements Driver {
             case ParamTypes.PARAM:
                 return context.params[param.name];
             case ParamTypes.SESSION:
-                throw new Error("@Session decorator is not supported by KoaDriver yet.");
-                // TODO return session
+                if (context.session === undefined) {
+                    throw new Error("Session is undefined! Is any Session-Middleware exist?");
```
pleerock: probably better message should be given. adding @19majkel94 for better ideas

test/functional/action-params.spec.ts
```diff
@@ -1,11 +1,13 @@
 import "reflect-metadata";
-import * as session from "express-session";
+const convert = require("koa-convert");
+const KoaSession = require("koa-session");
```
pleerock: move these blocks under 25 line to be consistent

src/ParamHandler.ts
```diff
@@ -104,7 +104,7 @@ export class ParamHandler {
     private parseValue(value: any, paramMetadata: ParamMetadata) {
         try {
             const parseValue = typeof value === "string" ? JSON.parse(value) : value;
-            if (paramMetadata.format !== Object && paramMetadata.format && this.driver.useClassTransformer) {
+            if (!(value instanceof paramMetadata.format) && paramMetadata.format !== Object && paramMetadata.format && this.driver.useClassTransformer) {
```
pleerock: ohhh this code gets messier... can you add comment why you added this line, also make sure to include test that coverts this scenario

src/decorator/params.ts
```diff
@@ -69,6 +69,32 @@ export function Param(name: string) {
 }
 
 /**
+ * This decorator allows to inject a State object to the controller action parameter.
+ * Applied to class method parameters.
+ *
+ * @param objectName The name of object stored in State
+ */
+export function State(objectName?: string) {
```
pleerock: TBH I didn't understand why this decorator is needed and what its relation to session. Can you please explain?

src/driver/KoaDriver.ts
19majkel94: Throwing Error here is a bad idea, it should be removed and just returned undefined or not. Checking if the param is empty is already [here](https://github.com/pleerock/routing-controllers/blob/master/src/ParamHandler.ts#L45).

src/decorator/params.ts
19majkel94: State is not a session, it's a recomended place for storing data for passing between middlewares:
https://github.com/koajs/koa/blob/master/docs/api/context.md#ctxstate

test/functional/action-params.spec.ts
mingyang91: Oh, this commit is used to test why the codecov / patch check can not be successful. I will remove it.

src/ParamHandler.ts
mingyang91: The state field can be assigned to an object of a class that does not need to be converted.
An error occurs if you execute plainToClass on the object.

README.md
```diff
@@ -415,8 +415,27 @@ savePost(@Session("user") user: User, @Body() post: Post) {
 }
 ```
 
-Routing-controllers uses [express-session][5] to handle session, so firstly you have to install it manually to use `@Session` decorator.
-This feature is not supported by koa driver yet.
+Routing-controllers Express uses [express-session][5] / Koa uses [koa-session][6] or [koa-generic-session][7] to handle session, so firstly you have to install it manually to use `@Session` decorator.
```
pleerock: add more details docs what needs to be installed

pleerock:
@19majkel94 please confirm this PR

src/driver/ExpressDriver.ts
```diff
@@ -175,7 +175,7 @@ export class ExpressDriver extends BaseDriver implements Driver {
                     return request.session;
                 }
             case ParamTypes.STATE:
-                throw new Error("@State decorators are not supported by ExpressDriver yet.");
+                return undefined;
```
19majkel94: Now you removed the good line 😉 
`throw new Error("@State decorators are not supported by ExpressDriver yet.");` should stay to tell user that it won't work with Express

test/functional/action-params.spec.ts
```diff
@@ -72,6 +78,46 @@ describe("action parameters", () => {
         // reset metadata args storage
         defaultMetadataArgsStorage().reset();
 
+
+        @Middleware()
+        class SessionMiddleware implements MiddlewareInterface {
```
19majkel94: `SessionMiddleware` should be moved to `/test/fakes/global-options/`

test/functional/action-params.spec.ts
```diff
@@ -72,6 +78,46 @@ describe("action parameters", () => {
...
+        }
+
+        class User {
+            public username: string;
+            public location: string;
+            public twitter: string;
+        }
+
+        @Middleware()
+        class SetStateMiddleware implements MiddlewareInterface {
```
19majkel94: `SetStateMiddleware` should be moved to `test/fakes/global-options/koa-middlewares`

test/functional/action-params.spec.ts
```diff
@@ -99,12 +99,20 @@ describe("action parameters", () => {
...
             public use (context: any, next: (err?: any) => Promise<any>): Promise<any> {
-                context.state = {
-                    username: "pleerock"
-                };
+                const user = new User();
+                user.username = "pleerock";
+                user.location = "Dushanbe, Tajikistan";
+                user.twitter = "https://twitter.com/pleerock";
+                context.state = user;
                 return next().then(() => { console.log("here3", context.state); });
```
19majkel94: `console.log` in tests are a bad idea, if that was for debug please remove it 😉

test/functional/action-params.spec.ts
mingyang91: This will cause some problems, `defaultMetadataArgsStorage().reset();` will clear all of the middleware.

19majkel94: I'm not sure, have you tried?
In other test suites the middlewares are defined with the test code so it can stay here were it is.
The `/test/fakes/` folder is used only for testing loading middlewares and error handlers from the directories.

test/functional/action-params.spec.ts
```diff
@@ -77,49 +73,13 @@ describe("action parameters", () => {
...
-                const user = new User();
-                user.username = "pleerock";
-                user.location = "Dushanbe, Tajikistan";
-                user.twitter = "https://twitter.com/pleerock";
-                context.state = user;
-                return next().then(() => { console.log("here3", context.state); });
-            }
-        }
 
+        const {SetStateMiddleware} = require("../fakes/global-options/koa-middlewares/SetStateMiddleware");
```
19majkel94: Does importing middlewares works even with `defaultMetadataArgsStorage().reset();`?
Why JS require instead of import?

src/driver/ExpressDriver.ts
```diff
@@ -175,6 +174,7 @@ export class ExpressDriver extends BaseDriver implements Driver {
                     return request.session;
                 }
             case ParamTypes.STATE:
+                console.warn("[routing-controllers] @State decorators are not supported by ExpressDriver yet.");
```
19majkel94: Please move back to throwing error like it's already in koa driver, no console!
```typescript
case ParamTypes.STATE:
    throw ner Error("@State decorators are not supported by ExpressDriver yet.");
```

test/functional/action-params.spec.ts
mingyang91: I have tried.
* Middleware define in `before` function inside ✅
* Middleware define in other file
    - file head `import`  ❌
    - `require` ✅
    - delete `defaultMetadataArgsStorage().reset()` ✅

mingyang91: I guess, this is because the `Middleware` in the `global-options` is imported with `options`.

https://github.com/pleerock/routing-controllers/blob/master/test/functional/load-from-directory.spec.ts#L72

19majkel94:
@pleerock looks like @mingyang91 have fixed all shortcomings, you can check and merge then

pleerock:
@19majkel94 @mingyang91 thank you guys!

PR #78, opened 2017-02-07
Added option for auto validate action params

19majkel94:
I have added integration with class-transformer-validator.
This lets to perform auto-validation of method params thanks to class-validator.
You don't have to manually parse and check received object, you can use validation decorators and let the magic do that work for you 😄 
And you don't have to remember to call validator function in each controller method, you just need to turn the option in server bootrstrap.


README.md
```diff
@@ -973,6 +972,61 @@ Learn more about class-transformer and how to handle more complex object constru
...
+npm install class-transform-validator --save
+```
+
+Then simply specify a `useParamValidator: true` option on application bootstrap:
+```typescript
+import "reflect-metadata";
+import { createExpressServer } from "routing-controllers";
+
+createExpressServer({
+    useParamValidator: true
```
pleerock: maybe `useValidator` is enough? or maybe `enableValidator` or `enableValidation`?

README.md
```diff
@@ -973,6 +972,61 @@ Learn more about class-transformer and how to handle more complex object constru
...
+}).listen(3000);
+```
+
+Now you need to define the class which type will be used in controller method params. Decorate the properties with appropriate validators.
```
pleerock: probably `with appropriate validation decorators`?

README.md
```diff
@@ -973,6 +972,61 @@ Learn more about class-transformer and how to handle more complex object constru
...
+```
+If you haven't used class-validator yet, you can learn how to use the decorators and handle more complex object validation [here][6].
+
+Now, if you have specified a class type, your action params will be not only an instance of that class (with the data sent by a user) but they will be validated too, so you don't have to worry about eg. incorrect e-mail or too short password and manual checks every property in controller method body.
+```typescript
+@Controller()
+export class UserController {
+
+    @Post("/login/")
+    login(@Body() user: User) {
```
pleerock: probably it also a good to have ability to control validation per-use-case, e.g. `@Body({ validate: false|true })`

README.md
```diff
@@ -973,6 +972,61 @@ Learn more about class-transformer and how to handle more complex object constru
 This behaviour is enabled by default.
 If you want to disable it simply pass `useClassTransformer: false` to createExpressServer method.
 
+## Auto validating action params
+
+Sometimes parsing a json object into instance of some class is not enough. 
+E.g. `class-transformer` doesn't check whether the property's types are correct, so you can get runtime error if you rely on TypeScript type safe. Also you may want to validate the object to check e.g. whether the password string is long enough or entered e-mail is correct.
+
+It can be done easily thanks to integration with [class-transformer-validator][7] which internally depends on [class-validator][6]. So first of all, you have to install the modules:
```
pleerock: This dependency graph looks complicated. Maybe we simplify things and make class-validator dependency of routing-controllers? And move all code from class-transform-validator into routing-controllers repository? What do you think about it?

README.md
```diff
@@ -973,6 +972,61 @@ Learn more about class-transformer and how to handle more complex object constru
 This behaviour is enabled by default.
 If you want to disable it simply pass `useClassTransformer: false` to createExpressServer method.
 
+## Auto validating action params
```
pleerock: Can `@Body` be called "action param"?

README.md
```diff
@@ -973,6 +972,61 @@ Learn more about class-transformer and how to handle more complex object constru
...
+    login(@Body() user: User) {
+        console.log(`${user.email} is for 100% sure a valid e-mail adress!`);
+        console.log(`${user.password.length} is for 100% sure 6 chars or more!`);
+    }
+
+}
+```
+If the param doesn't satisfy the requirements defined by class-validator decorators, an error will be throwed and captured by routing-controller, so the client will receive 400 Bad Request and JSON with nice detailed [Validation errors](https://github.com/pleerock/class-validator#validation-errors) array.
+
+If you need special options for validation (groups, skiping missing properties, etc.) or transforming (groups, excluding prefixes, versions, etc.) you can pass them as global config as `paramValidatorOptions` in createExpressServer method or as a local setting for method parameter - `@Body({paramValidatorOptions: localOptions})`.
```
pleerock: maybe we can call `paramValidatorOptions` simply `validatorOptions` ?

src/ParamHandler.ts
```diff
@@ -104,9 +106,19 @@ export class ParamHandler {
     private parseValue(value: any, paramMetadata: ParamMetadata) {
```
pleerock: we'll need to refactor this block later or, because its very hard to understand what is going on there

src/ParamHandler.ts
```diff
@@ -104,9 +106,19 @@ export class ParamHandler {
...
-            if (paramMetadata.format !== Object && paramMetadata.format && this.driver.useClassTransformer) {
-                const options = paramMetadata.classTransformOptions || this.driver.plainToClassTransformOptions;
-                return plainToClass(paramMetadata.format, parseValue, options);
+            if (paramMetadata.format !== Object && paramMetadata.format) {
+                if (this.driver.useParamValidator) {
+                    const options = paramMetadata.paramValidatorOptions || this.driver.paramValidatorOptions;
+                    return transformAndValidate(paramMetadata.format, parseValue, options)
+                        .catch(err => {
+                            const error: any = new BadRequestError(`Invalid ${paramMetadata.type}, check 'errors' property for more info`);
+                            error.errors = err;
```
pleerock: err inside errors which also has errors. This looks too nesty. Maybe its better to return same validation object, just change its message and httpCode to match 400 code?

src/RoutingControllersOptions.ts
```diff
@@ -1,4 +1,5 @@
 import {ClassTransformOptions} from "class-transformer";
+import {TransformValdiationOptions} from "class-transformer-validator";
```
pleerock: if users will install `routing-controllers` without `class-transformer-validator` then their typescript compiler will always give error that it can't find `class-transformer-validator` package. So, using this expression you force users to always install `class-transformer-validator`

README.md
19majkel94: Action param -> controller method parameter, right?

src/ParamHandler.ts
19majkel94: `transformAndValidate()` throws `ValidationError[]` so this array can't be just returned as JSON and you can't just add the message and httpCode properties.
I added the `catch` block to transform the `ValidationError[]` to the `BadRequestError` which can be easily handled by the default error handler.
But I will refactor the names inside this block to make it more readable 👍

README.md
19majkel94: We could simplify it and make `class-validator` dependency of routing-controllers. If so, also `class-transformer-validator` could be the dependency as it's really small lib and has only peer dependency on `class-validator` and `class-transformer`.

README.md
19majkel94: Sure! I think I just wanted to avoid "decorate with decorators" 😉

README.md
pleerock: if `class-transformer-validator` is so small then maybe its better to have its code inside `routing-controllers` library?

19majkel94: I would like to use it also in my other projects and not to be forced to download the whole `routing-controllers`.
If you are afraid because it's not your npm module, etc., I could lock the dependency version to 0.1.1 😉

src/RoutingControllersOptions.ts
19majkel94: You are right, I checked that now. There are two options:
- use `skipLibCheck` option for tsc
- as it was said before, we could make this libs the dependency
- make a JS require and forget about typings for this 😉

README.md
pleerock: > not your npm module

I don't care about it, this package is for open-source, not for me. I do care about complexity and maintainability. I have lot of packages and know about maintainability costs when you have lot of packages to maintain, and this small extra package to maintain does not cost it. What exactly this package does? How can you use it without routing-controllers? Isn't it tightly bind to routing-controllers? Or is there anything non-specific to routing-controllers? Maybe routing-controllers does not need it at all? And just have only core really needed from this package inside routing-controllers, and you can use your package with extra specific features you need?

README.md
19majkel94: ok, changed to `enableValidation`

README.md
19majkel94: good point, I wanted to do this and forget about this 😢 done in 28b7adf

README.md
19majkel94: ok, changed name to `validatorOptions` 😉

src/ParamHandler.ts
19majkel94: 59801f9 made this code more readable? or you have an other idea how the error handler should return correct http code and nice JSON with info?

README.md
pleerock: just looked at `class-transformer-validator` and realize that its not related to routing-controllers and can be used just with validator+class-transformer. But inside it just calls two methods of two libraries in sequence and provide a single function to use, right? If so, then do routing-controllers need it? Probably no, because it does not need such "facade function" because its used only once and is internal. It simply can call two methods of two libraries in sequence by its own. So, now the question is: Does it worth to bind routing-controllers to extra dependency that makes a small amount of work that routing-controllers can do on its own? Answers is in my opinion is no. But this tool can be useful for those who are using it just with validator+class-transfromer, without routing-controllers, as you told.

src/ParamHandler.ts
pleerock: its better now that you add type of data in catch block. However I think `error.errors` was better then `error.details`

README.md
19majkel94: `class-transformer-validator` isn't tightly coupled with `routing-controllers`, it can work with anything, eg. with pure express.js:
```typescript
async (req, res) => {
    try {
        const post = await transformAndValidate(Post, req.body);
        await Post.persist(post);
        res.send(`Post ${post.getFullName()} has been added!`);
    } catch (error) {
        res.status(400).json(error);
    }
}
```
Or even for loading config JSON:
```typescript
try {
    const jsonString = await fs.readFileAsync("./files/config.json", "utf8");
    const config = await transformAndValidate(Config, jsonString);
} catch (error) {
    console.err(error);
}
```

Of course, it's just a simple wrapper which "simply call two methods of two libraries in sequence" and you could do this by its own. But if you try to make the code clean and keep SOLID, due to Single responsibility principle you should move that part of code into a new function (and make some tests?) which already has been done in `class-transformer-validator` 😉

It's not a 100KB library with 50 deps from which we need 1 function. I think that we should DRY and don't make copy-paste of every lib just to avoid dependency.

pleerock: > class-transformer-validator isn't tightly coupled with routing-controllers, it can work with anything, eg. with pure express.js:

yes Ive got this.


I hate dependencies. Dependencies makes software more complex and harder to maintain. Especially small dependencies. When there is no real need to have extra dependency and maintainability cost for such a small functionality that can take just few lines of code there.

pleerock:
@19majkel94 did we come to some conclusion with this PR?)

19majkel94:
@mingyang91 I merged your changes but now I have test error - `@State` is not working for koa if we use class type in action parameter (`@State() user: User`).

And there is an error in your `SetStateMiddleware`:
```
TypeError: next is not a function
    at SetStateMiddleware.use (\routing-controllers\build\es5\test\fakes\global-options\koa-middlewares\SetStateMiddleware.js:19:16)
```
Could you check what's going on?

19majkel94:
@pleerock I don't know what is the conclusion 😉 It's your decision weather you want to keep this PR or "inline" the integration with class-validator by yourself.

mingyang91:
The real error message is
```
  1) action parameters @State should return a value from state asserting port 3002:
     AssertionError: expected status code 404 to equal 200
      at build/es5/test/functional/action-params.spec.js:496:36
      at _fulfilled (node_modules/q/q.js:834:54)
      at self.promiseDispatch.done (node_modules/q/q.js:863:30)
      at Promise.promise.promiseDispatch (node_modules/q/q.js:796:13)
      at node_modules/q/q.js:604:44
      at runSingle (node_modules/q/q.js:137:13)
      at flush (node_modules/q/q.js:125:13)
      at _combinedTickCallback (internal/process/next_tick.js:67:7)
      at process._tickDomainCallback (internal/process/next_tick.js:122:9)
```

I think the state parameter is `null` because test case is 
```
            @Get("/state")
            @UseBefore(SetStateMiddleware)
            @JsonResponse()
            getState(@State() state: User) {
                return state;
            }
```
So `HTTP Status` is 404

src/ParamHandler.ts
```diff
@@ -106,10 +110,20 @@ export class ParamHandler {
...
+                        .catch((validationErrors: ValidationError[]) => {
+                            const error: any = new BadRequestError(`Invalid ${paramMetadata.type}, check 'details' property for more info.`);
+                            error.details = validationErrors;
+                            throw error;
+                        });
+                // If value is already an instance of target class, then skip the plain to class step
+                } else if (!(value instanceof paramMetadata.format) && this.driver.useClassTransformer) {
+                    const options = paramMetadata.classTransformOptions || this.driver.plainToClassTransformOptions;
+                    return plainToClass(paramMetadata.format, parseValue, options);
+                }
```
mingyang91: This will return undefined, may need an else block.

19majkel94:
@pleerock Now everything looks ok and you should be able to merge this PR 😉

pleerock:
@19majkel94 thank you so much! This is a really awesome new feature

pleerock:
released in `0.6.10`

PR #82, opened 2017-02-17
swagger.json controller based on metadata

awaragi:
This is a quick draft of swagger definition generator based on currently collected metadata
remains to do
- basic info section
- add @Api and other tags as per swagger annotation project (https://github.com/swagger-api/swagger-core/wiki/Annotations-1.5.X)
- add documentation

Please let me know if this is of interest. The swagger specific annotations can be done separately but I believe the framework can greatly benefit from swagger definition right next to controllers definitions all bundled together.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,5 @@ node_modules/
 typings/
 coverage/
 npm-debug.log
+*.iml
+/.idea/
--- a/src/RoutingControllersOptions.ts
+++ b/src/RoutingControllersOptions.ts
@@ -1,5 +1,6 @@
 import {ClassTransformOptions} from "class-transformer";
 import {ValidationOptions} from "class-validator";
+import {SwaggerOptions} from "./swagger/SwaggerDefinition";
 
 /**
  * Routing controller initialization options.
@@ -88,5 +89,9 @@ export interface RoutingControllersOptions {
      * Route prefix. eg '/api'
      */
     routePrefix?: string;
-    
+
+    /**
+     * Swagger options. By default disabled.
+     */
+    swagger?: SwaggerOptions;
 }
--- a/src/index.ts
+++ b/src/index.ts
@@ -6,6 +6,7 @@ import {KoaDriver} from "./driver/KoaDriver";
 import {Driver} from "./driver/Driver";
 import {getFromContainer} from "./container";
 import {RoutingControllersOptions} from "./RoutingControllersOptions";
+import {SwaggerDefinition} from "./swagger/SwaggerDefinition";
 
 // -------------------------------------------------------------------------
 // Main Functions
@@ -98,6 +99,12 @@ function createExecutor(driver: Driver, options: RoutingControllersOptions): voi
 
     if (options.routePrefix !== undefined)
         driver.routePrefix = options.routePrefix;
+
+    if (options && options.swagger && options.swagger.enabled) {
+          new SwaggerDefinition(options.swagger)
+              .bootstrap(options.routePrefix || "/")
+              .controller();
+    }
 
     // next create a controller executor
     new RoutingControllerExecutor(driver)
--- a/src/swagger/SwaggerDefinition.ts
+++ b/src/swagger/SwaggerDefinition.ts
@@ -1,0 +1,165 @@
+import {JsonController, Controller} from "../decorator/controllers";
+import {Get} from "../decorator/methods";
+import {MetadataBuilder} from "../metadata-builder/MetadataBuilder";
+import {UseMetadata} from "../metadata/UseMetadata";
+import {ParamTypes} from "../metadata/types/ParamTypes";
+import {JsonResponse, Header} from "../decorator/decorators";
+import {Res} from "../decorator/params";
+
+export interface SwaggerOptions {
+    /**
+     * Defaults to false
+     */
+    enabled?: boolean;
+    /**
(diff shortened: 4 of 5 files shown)


awaragi:
@pleerock Hi, any chance for quick feedback if this is of interest to the projet before I invest more time in it? Thanks in advance.

pleerock:
Hi @awaragi sorry for the delayed response I was really busy last few days and this PR was required me to go deeper then just a quick look. I already like what you did and for sure there is a group of people who are interested in having swagger for routing-controllers. But I'm not sure if its a good choice to have it as a part of routing-controllers, or it should be a separate library. Because it bring lot of decorators which aren't needed for every users I mostly think it should be a separate library which however is deeply integrated with routing-controllers. What do you think about it? Adding @19majkel94

19majkel94:
Right, I think that it should be moved to the separate plugin as Typescript allows for declarations mergin, so it's easy to extend interface `RoutingControllersOptions` to add new property.

The only problem is with [this four lines](/pleerock/routing-controllers/pull/82/files#diff-f41e9d04a45c83f3b6f6e630f10117feR103), it has to stay in `routing-controller` basecode but with `require()` the swagger-plugin inside the if block:

```typescript
if (options && options.swagger && options.swagger.enabled) {
    const SwagerPlugin = require("routing-controllers-swagger-ext");
    new SwagerPlugin.SwaggerDefinition(options.swagger)
        .bootstrap(options.routePrefix || "/")
        .controller();
}
```
Then what about tests, the extesion should be in dev-deps and have some test in this codebase?

awaragi:
Hi Guys, thank you for your response. A plugin is one way to go and I have considered originally. My reasons for including swagger generator builtin are as follows:
- routing controllers is still gaining momentum, more builtin features = easier sell to architecture group
- expressjs has no builtin swagger generator and having separate and external plugins goes against my believe that documentation should be as close as possible to code for it to remain up to date.
- if you do nothing but enable swagger in createExpressServer options then you will a get fully working documentation of the end points minus the typescript interfaces descriptions (which are impossible in typescript as interfaces are lost after transpilation). Additional tags (thinking @Api, @ApiParam and @ApiModel only) only add more content.

Ultimatey, this package and decision are yours but I believe builtin out of the box swagger.json will be a great feature.

Please let me know what you think.

19majkel94:
> "minus the typescript interfaces descriptions (which are impossible in typescript as interfaces are lost after transpilation)"

So maybe it would be better if `swagger.json` were generated at build-time? You could use TypeScript Language Service to get all info about types (including interfaces), and swagger.json could be then served as static resource?

awaragi:
I think that would be a great approach. The output would be a simple json object that can be loaded by any controller the developer desires. The generation would be posttsc task (npm speaking) thus always executed and provide up to date documentation.

I didn't consider this approach since I didn't know about it! Would you be willing to guide me a little bit with that? It obviously needs to scan for all the routing-controllers decorators, type interfaces for the parameters and finally the couple of additional decorators for swagger specific documentation.

Remains the question about whether this should be a separate package or part of routing-controllers. My arguments for inclusion remain but the approach changes.

pleerock:
I think its more suitable as a separate package which will have very smooth integration with routing-controllers.

> routing controllers is still gaining momentum, more builtin features = easier sell to architecture group

We can left it as a separate package, but add a good section in readme that swagger is supported and provide an examples, but left a note that before user need to do `npm i routing-controllers-swagger` for example. Effect on selling will be same.

The problem with all-in-one is that not everybody need specific features (more over if they arent small) and this is loads the framework. Plus there will be lot of new decorators which may conflicts with other libraries like `@Api` which is solvable problem, but is not pleasant.

> expressjs has no builtin swagger generator and having separate and external plugins goes against my believe that documentation should be as close as possible to code for it to remain up to date.

that's a price that comes with modularity. Modularity brings lot of good things like extensibility, flexibility and developing by teams in isolation, but also brings cost of extra maintainability. 

In the future I'm planning to move all repositories (typedi, class-validator, class-transformer, routing-controllers) under [typestack](https://github.com/typestack) organization and create a single framework which will have them all merged and provided a convenient way.

awaragi:
Okay. based on your recommendations, I will create a separate package for swagger.json generator using base annotations and type/interfaces for models. The TSOA package has a similar concept so I will get inspiration from it (perhaps using jsDoc instead of additional annotations).

pleerock:
I think you'll still need additional decorators because not everything you can cover with jsdoc or typescript language service. So the best in my opinion to use metadata provided by routing-controllers + metadata provided in decorators + rely on jsdoc or ts language service if its convenient for end user to do so + is reliable. If you need help with separate library let me know what I can help with

amcdnl:
@awaragi - I've started a similar project to this aimed to have direct swagger support out of the box. We just did a 1.0 release and your work here is quite amazing. I'm planning to use it as my starting place. Would love any contributions! https://github.com/swimlane/trafficlight

Sorry for solicitation @pleerock ;P

et:
Hey @awaragi - I was wondering if you ever revisited this topic. I just started using `routing-controllers` and would love to try an implementation for swagger docs generation.

19majkel94:
@et You can try use Nest, it has support for swagger docs 😉

awaragi:
I just took a look at Nest and I have to agree with @19majkel94 that it might be a good alternative.

PR #86, opened 2017-02-23
Creating test case for issue with class-transformer

alexthebuilder:
@pleerock here is the example I have on routing controllers for my issue with the class transformer. If you locally link the change in class-transformer it will fix the test issues this change has caused.

### DIFF
--- a/test/functional/class-transformer-options.ts
+++ b/test/functional/class-transformer-options.ts
@@ -9,7 +9,7 @@ import {
 } from "../../src/index";
 import {QueryParam} from "../../src/decorator/params";
 import {assertRequest} from "./test-utils";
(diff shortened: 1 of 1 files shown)


pleerock:
@alexthebuilder actually I was talking to create a test case in the class transformer library, not here. It does not mean that you should install routing-controllers in the class-transformer library. It means that this issue should not specific to routing-controllers and it can be represented in isolation

alexthebuilder:
@pleerock I understand what you are asking but the level of effort to isolate the issue due to the way its used in this the routing-controller is a bit of work. I was hoping this test case would show you the issue. I'll put some effort to try and comb through the code here to isolate why the other module is breaking.

PR #90, opened 2017-03-03
Implemented Koa file upload

jtmthf:
Uses koa-multer instead of koa-router-multer as it seems more popular. All references and tests should be up to date, but let me know if I missed anything.


pleerock:
Thank you!

PR #101, opened 2017-03-27
Reduce the number of get access calls and improve performance.

mingyang91:
Because each call get accessor, will be obtained from the container, which is unnecessary.


src/driver/ExpressDriver.ts
```diff
@@ -61,12 +61,13 @@ export class ExpressDriver extends BaseDriver implements Driver {
      * Registers middleware that run before controller actions.
      */
     registerMiddleware(middleware: MiddlewareMetadata): void {
-        if (!middleware.instance.use)
+        const instance = middleware.instance;
```
pleerock: are such changes really increases performance so much? Also good to have a benchmark results

pleerock:
closing it because of no answer from author

PR #103, opened 2017-04-01
Fix not working validationOptions and it's type

### LINKED ISSUE #102: ValidationOptions has no property skipMissingProperties?
I am using routing-controllers with typescript.
I  want to use skipMissingProperties when I use createExpressServer method.
But, skipMissingProperties  does not in  ValidationOptions ，as fllows.
```
/**
 * Options used to pass to validation decorators.
 */
export interface ValidationOptions {
    /**
     * Specifies if validated value is an array and each of its item must be validated.
     */
    each?: boolean;
    /**
     * Error message used to be used on validation fail.
     * You can use "$value" to use value that was failed by validation.
     * You can use "$constraint1" and "$constraint2" keys in the message string,
     * and they will be replaced with constraint values if they exist.
     * Message can be either string, either a function that returns a string.
     * Second option allows to use values and custom messages depend of them.
     */
    message?: string | ((value?: any, constraint1?: any, constraint2?: any) => string);
    /**
     * Validation groups used for this validation.
     */
    groups?: string[];
    /**
     * Indicates if validation must be performed always, no matter of validation groups used.
     */
    always?: boolean;
}
```

How to solve this problem ?  thanks

19majkel94:
Fixed #102.

### DIFF
--- a/src/RoutingControllersOptions.ts
+++ b/src/RoutingControllersOptions.ts
@@ -1,5 +1,5 @@
 import {ClassTransformOptions} from "class-transformer";
-import {ValidationOptions} from "class-validator";
+import {ValidatorOptions} from "class-validator";
 
 /**
  * Routing controller initialization options.
@@ -55,7 +55,7 @@ export interface RoutingControllersOptions {
     /**
      * Global class-validator options passed during validate operation.
      */
-    validationOptions?: ValidationOptions;
+    validationOptions?: ValidatorOptions;
 
     /**
      * Global class transformer options passed to class-transformer during classToPlain operation.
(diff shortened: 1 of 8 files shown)


src/RoutingControllersOptions.ts
```diff
@@ -55,7 +55,7 @@ export interface RoutingControllersOptions {
     /**
      * Global class-validator options passed during validate operation.
      */
-    validationOptions?: ValidationOptions;
+    validationOptions?: ValidatorOptions;
```
pleerock: I don't like `ValidationOptions` and `ValidationOptions` in class-validator package. Good to rename then however I didnt find a good name for both when was creating them

pleerock:
thank you)

Issue #108, opened 2017-04-05
Ugly @MiddlewareGlobalAfter and @MiddlewareGlobalBefore


pleerock:
I really don't like `@MiddlewareGlobalAfter` and `@MiddlewareGlobalBefore` names. I never did. What if we find a better name for those decorators or deprecate them and use simply:

```ts
@Middleware({ global: true, type: "after"|"before" })
```

instead? @19majkel94

19majkel94:
[+1 👍](https://www.youtube.com/watch?v=-uRi_xo92VEe)

Also we have to split middleware interface for koa and express.

pleerock:
yes agree

pleerock:
fixed in 0.7.0

Issue #96, opened 2017-03-09
Can't set limit in body parser


snirs90:
I'm trying to set a "limit" with the body parser library (using express) and it is not working due to that the "routing-controller" loads the `bodyParser.json()` with empty options.

example:
```
this.app = createExpressServer({
            defaultErrorHandler: false,
            routePrefix: '/api',
            controllers: [ appPath + '/src/controllers/*.js'],
            middlewares: [ appPath + "/src/middlewares/*.js"]
        });
this.app.use(bodyParser.json({limit: '20mb'}));
```

This is not working because the bodyParser is called first with empty options by "JsonController" or "JsonResponse" decorators.

snirs90:
Also I've noticed, that if I'm not using `@JsonController` or `@JsonResponse`, when I'm trying to use to populate a `@Body` param with a class it is not working. example:
`post(@Body() candidate: CandidateForm)`

If the controller class doesn't have `@JsonController` or the method doesn't has `@JsonResponse`, the `candidate` variable is not populated at all and remain empty.

snirs90:
So I've manage to overcome this by configuring my own Express server and then using `useExpressServer`.
At the moment it is not possible to work with `createExpressServer` if you need to configure the body-parser json method.

@pleerock If you think that need to add the ability to configure it via `createExpressServer` also, I can help you out and add this feature.

Thanks,

19majkel94:
It is simple - if you don't have a special requirements, you use `createExpressServer` but if you need some middlewares you use `useExpressServer` which acts as a middleware and attach it into `app.use` chain.

Also you can use [global middlewares](https://github.com/pleerock/routing-controllers#global-middlewares) and configure there your 3rd party middlewares to work with routing-controllers.

pleerock:
@19majkel94 author is talking about [these lines](https://github.com/pleerock/routing-controllers/blob/0b9d319f74d14de9638b5f54e03a286886057c3a/src/driver/ExpressDriver.ts#L126-L126). Right now body-parser will be enabled anyway without configuration.

If we configure body parser on a global level, wont these lines of code override body parser of the currently executed route with empty configuration?

19majkel94:
@pleerock `ParamOptions` (and `ParamMetadataArgs`) have property `parseJson` - shouldn't that prevent from parsing the body by `body-parser`?

[...]

Issue #75, opened 2017-02-05
Support for custom param decorators


19majkel94:
I have seen that many people fork this repo to add custom decorators for exporting object from request object and inject it in the controllers methods params. It's a bit pain to modify switch-case in core of routing-controllers every time we need e.g. user object, jwt token or sth else.

I like the way that `class-validator` let us define our custom validation decorators:
https://github.com/pleerock/class-validator#custom-validation-decorators

It would be great if we could write custom decorators function to extract objects from request rather than forking and patching `routing-controllers`. It could be done the same way - the decorator function should call `registerParamHandler` with config object with required data (like in `ParamMetadataArgs`) and handler field - the function to call to extract:

```typescript
export function CustomDecorator(customParam?: customType) {
    return function (object: Object, methodName: string, index: number) {
        let format = (Reflect as any).getMetadata("design:paramtypes", object, methodName)[index];
        registerParamHandler({
            reflectedType: format,
            isRequired: true,
            classTransformOptions: undefined,
            // other settings here
            handler: (req: Request) => {
                // do the extract operation here
                if (customParam == someValue) {
                    return req.property.more.arr[someValue];
                } else {
                    return req.property.more.value;
                }
            }
        });
    };
}
```
The handlers could be pushed to array of handlers and in core of routing-controllers, after matching standard param types it could traverse throught this array, call the handlers and do the transform magic 😀

pleerock:
Here is how I do it right now in my project:

```typescript

/**
 * Extends routing-controllers and socket-controllers functionality and allows to inject currently
 * authorized user (SessionUser) into controller's actions.
 */
export function InjectSessionUser(options?: { required?: boolean }) {
    return function(object: Object, methodName: string, index: number) {
        const reflectedType = (Reflect as any).getMetadata("design:paramtypes", object, methodName)[index];

        // add special param handler to routing-controllers library
        defaultRoutingMetadataArgsStorage().params.push({
            target: object.constructor,
            method: methodName,
            index: index,
            type: RoutingParamTypes.CUSTOM_CONVERTER,
            reflectedType: reflectedType,
            parseJson: false,
            isRequired: false,
            format: reflectedType,
            transform: (value, request) => handleUserAuthorization(request.headers["authorization"], options)
        });

    };
}
```

Yes, I agree that we need provide an easy way to create user's custom decorators like in class-validator and make it even easier.

19majkel94:
[...]

Issue #57, opened 2016-12-27
Sending raw data is impossible.


reflog:
Hi!
Thanks for an awesome project! It make writing services in TS a real pleasure.
One bug I came across - even for actions that are marked with a different @ContentType and not inside @JsonController - the result cannot be binary, for example, I cannot send image file.
The reason for it is the following code in ExpressDriver:
```
 } else {
                    response.send(String(result));
                }
```

Basically, even if a Buffer is returned - it's converted to string. It's not a correct behavior. At least some sort of escape hatch is needed. 

What would you recommend?

awaragi:
I had to return images. Easiest thing is use @Res() to inject response object and write directly through it

reflog:
@awaragi - do you have a code snippet to demonstrate, maybe?

awaragi:
@Get(":I'd") get(@Param("id") id: string, @Res() response: any) {
        return this.documentDAO.get(id).then((data) => {
            if (data) {
                response.append("Content-Type", data.mimeType);
                response.end(data.file.buffer, 'UTF-8');
            } else {
                response.status(404);
                // 404 handler will take care of message content
            }
        }).catch((error) => {
            response.status(500);
            return Status.failure(error.message);
        })
    }

pleerock:
this should be fixed in 0.7.0

PR #119, opened 2017-04-24
0.7.0

pleerock:


pleerock:
@19majkel94 you need to review this one.... 😮  
Please don't kill me for the changes I made 🙏  
Fixes #108 #96 #75 #57 #54 #21 #24  🐛

I'll also need to update the docs.

Also alternative names for `ActionProperties`, `authorizationChecker` and `Authorized` are welcomed. Also adding @slavafomin

src/RoutingControllers.ts
```diff
@@ -55,29 +64,25 @@ export class RoutingControllers {
     /**
      * Registers post-execution middlewares in the driver.
```
19majkel94: Is the comment out of date as now the type is `before` or `after`?

src/ActionParameterHandler.ts
```diff
@@ -74,7 +73,7 @@ export class ActionParameterHandler {
     // Private Methods
```
19majkel94: The section comment is out of date now.

README.md
```diff
@@ -69,47 +69,37 @@ You can use routing-controllers with [express.js][1] or [koa.js][2].
...
-    `npm install koa@next koa-router@next koa-bodyparser@next koa-multer --save`
+    `npm install koa koa-router koa-bodyparser koa-multer --save`
 
-    Optionally you can also install its [typings](https://github.com/typings/typings):
+    Optionally you can also install their typings:
 
-    `typings install dt~koa --save --global`
+    `typings install @types/koa --save --global`
     
 5. Its important to set these options in `tsconfig.json` file of your project:
```
19majkel94: You have skipped `4.` - use `1.` as guthub markdown will transform it to auto enumerated list.

src/decorator/Session.ts
```diff
@@ -0,0 +1,20 @@
+import {defaultMetadataArgsStorage} from "../metadata-builder/MetadataArgsStorage";
+
+/**
+ * Injects a Session object to the controller action parameter.
+ * Must be applied on a controller action parameter.
+ */
+export function Session(objectName?: string): Function {
```
19majkel94: Is the return type necessary there? Isn't type inference enough?

README.md
```diff
@@ -73,40 +69,29 @@ You can use routing-controllers with [express.js][1] or [koa.js][2].
...
     **b. If you want to use routing-controllers with *koa 2*, then install it and all required dependencies:**
 
-    `npm install koa@next koa-router@next koa-bodyparser@next koa-multer --save`
+    `npm install koa koa-router koa-bodyparser koa-multer --save`
 
-    Optionally you can also install its [typings](https://github.com/typings/typings):
+    Optionally you can also install their typings:
 
-    `typings install dt~koa --save --global`
+    `typings install @types/koa --save --global`
```
19majkel94: npm install 😆

pleerock: 😆

README.md
pleerock: yes I didnt worked on readme yet, will completely update it tomorrow and let you know once again

src/decorator/Session.ts
pleerock: just wanted to do it as an convention - to always put return types...

19majkel94:
Oh, I haven't check the last commit from an hour ago. And I don't like the idea of current user and authorization checker 😞 

All of earlier commits was to simplify things, remove dead code or rarely used functionality, and now you introduce this. Authorization checker works like a regular middleware but with hardcoded support in core (which is simple and unnecessary). Current user decorator also work's like any custom decorator, like the `JWT` that I've done in 5 minutes for my project:

```typescript
/**
 * Extends routing-controllers and socket-controllers functionality
 * and allows to inject object stored in JWT token into controller's actions.
 *
 * @param {string} propertyName name of the request property where express-jwt place JWT token object
 */
export function JWT(propertyName = "user") {
    return (object: object, methodName: string, index: number) => {
        const reflectedType = (Reflect as any).getMetadata("design:paramtypes", object, methodName)[index];

        // add special param handler to routing-controllers library
        defaultMetadataArgsStorage().params.push({
            target: object.constructor,
            method: methodName,
            index,
            type: ParamTypes.CUSTOM_CONVERTER,
            reflectedType,
            parseJson: false,
            isRequired: false,
            format: reflectedType,
            transform: (_, req) => req[propertyName]
        });
    };
}
```
It adds unnecessary overhead to the codebase like the `interceptor` which you removed :wink: I would think twice if it's really needed - maybe you can convince me with better example than this in `samples` directory?

I would like to do the task of custom decorator support (developer-friendly API + changes in param decorator handling - no switch-case in core, etc.) but it's not breaking change so it can be done later. It should solve the `current user` decorator problem 😉

19majkel94:
I would also consider one breaking change - remove `@JsonController` and use `@Controller({ json: true })` instead, like with `@Middleware`. What do you think?

gulpfile.js
```diff
@@ -1 +0,0 @@
-eval(require("typescript").transpile(require("fs").readFileSync("./gulpfile.ts").toString()));
```
marshall007: @pleerock how does this work now? Does `gulp-cli` support `gulpfile.ts` files out of the box now?

pleerock: now I use `ts-node` instead

pleerock:
> I would also consider one breaking change - remove @JsonController and use @Controller({ json: true }) instead, like with @Middleware.

@19majkel94 I also really really hate `@JsonController` name, however `@Controller({ json: true })` looks uglier to me.

> I would consider changes in param decorator handling - no switch-case in core, but all logic should be in value or transform function

I was thinking about this, but decided not to touch those things for now

> It adds unnecessary overhead to the codebase like the interceptor which you removed

necessary for some people, unnecessary for others... Im still not sure if interceptor removal was a good idea. Sometimes that really needed functionality.

`@Authorize` and `@CurrentUser` are commonly things, almost each routing-controllers project should have them, I think this way we introduce a "convention" for decorator names plus simplify their implementation for users

pleerock:
@19majkel94 maybe we shall enable class-validation by default? Because we won't loose something if it will be until user uses class-validator decorators, am I right?

19majkel94:
> I was thinking about this, but decided not to touch those things for now

It would really think about it as it really simplifies and unifies the logic, and also makes the creation of custom decorators really easy. BTW, `registerParamDecorator` can be implemented as a function which return a decorator function:
```typescript
export function UserFromSession(options?: { required?: boolean }) {
    return registerParamDecorator({
        required: options && options.required ? true : false,
        value: actionProperties => {
            const token = actionProperties.request.headers["authorization"];
            // return database.findUserByToken(token);
            return new User(1, "Johny", "Cage");
        }
    });
}

export function registerParamDecorator(options) {
    return function(object: Object, method: string, index: number) {
         _registerParamDecorator(object, method, index, options);
    }
}
```

19majkel94:
> maybe we shall enable class-validation by default? Because we won't loose something if it will be until user uses class-validator decorators, am I right?

Yeah, it should be turned on by default but I haven't done this as it might be breaking change (someone can decorate their classes but validate it manually only on some actions). Now we have major release so do the magic 😆

19majkel94:
> @Authorize and @CurrentUser are commonly things, almost each routing-controllers project should have them, 

Hmmm... `@CurrentUser`... I don't know, people use middlewares like `express-session` or `express-jwt` do to the logic which just attach object to `req` object (like `req.session` or `req.user`). Decorators should only be responsible of extracting a property to inject it in action's param, not to connect with database (get connection to db from container or singleton? we have DI in middlewares for this). 

If you think about commonly things, every backend project use authorization - sessions or JWT. I've added session decorator so you should also add JWT decorator (sample is above already).

Authorization checker feature... I think everyone do this in other way, it's not as simple as `@Body` decorator. Your solution is set globally, I would rather see this as an built-in middleware factory which users has to configure and use explicitly per route or controller.

19majkel94:
I also thought about support for declaring middlewares that should be used before param decorators do the extraction. Like you create `JWT` decorator and configure `express-jwt` for it. We could then remove hardcode depenedency for `@Body` and `multer`, etc.

And in major release we have planed the support for middleware functions in `@UseXXX` decorators, not only classes with `use` function.

pleerock:
> registerParamDecorator can be implemented as a function which return a decorator function:

awesome idea, will do that

pleerock:
updated readme and changelogs, I feel I miss something in changelogs...

pleerock:
adding @idchlife

idchlife:
@pleerock can I help somehow?

If you allow me in discussion...

About @CurrentUser and @Authorize

My opinion: @Authorize sounds like action should authorize user and then allow/disallow access. Also authorization is about rights-roles check, authentication on the other hand is about user existence in the application and his/hers authentication status (anonymous, authenticated)

@CurrentUser is a nice decorator!
But wait, why if we have req.user?
It could do some magic for variety express/koa/something else, but developer should know where to search user in chosen framework.

About @JsonController - why do you both hate it? It looks great! If we would have responses in our actions (but we do not, we have res.status, res.send, res.json etc) we would have something like new Response/new JsonResponse (you do get where I got them from, yes?)

But Json is a common format to give responses and get requests in. I'm using it and it really does the trick. Giving options to @Controller will make it look really uglier :(

Who knows, maybe somebody really wants XmlController in the future :D

pleerock:
@idchlife

> @Authorize sounds like action should authorize user and then allow/disallow access. Also authorization is about rights-roles check, authentication on the other hand is about user existence in the application and his/hers authentication status (anonymous, authenticated)

correct.

> @CurrentUser is a nice decorator!
But wait, why if we have req.user?

because this:

```ts
action(@CurrentUser() currentUser: User)
```

is more elegant then:

```ts
action(@Req() req: Request)
const currentUser = req.user;
```

plus `@CurrentUser` provides convenient `{ required: true }` option.

> maybe somebody really wants XmlController in the future :D

only if we go back to the future :)

idchlife:
@pleerock I agree about elegance. But I discovered that my actions are becoming bloated already with @Body, @Req, @Res, @Session etc. I think that's everyones personal choice of course. Having @CurrentUser is a nice touch to existing parameters decorators.

{ required: true } will throw error "required argument/user" if user does not exist, right? But not UnauthorizedError or smthn?

19majkel94:
> because this:
```typescript
action(@Req() req: Request)
const currentUser = req.user;
```

Welcome in ES6 destruction world 😉 

```typescript
action(@Req() { user }: Request)
```
I've used object destruction also on response object:
```typescript
action(@Param("id") id: number, @Res() { sendFile }: express.Response)
```
But it complicates unit testing of controllers - you have to pass whole `Response` object or assert as any. So the more type-safe way is this weird thing:
```typescript
@Res() { sendFile }: { sendFile: express.Response["sendFile"] })
```
I would love to see the parametrized version of `@Req` and `@Res` decorator, which should work like a `@Session("property")` decorator. So I would do this:
```typescript
@Res("sendFile") sendFile: express.Response["sendFile"])
```
With this developers could do the `req.user` trick without weird syntax or custom decorators:
```typescript
action(@Req("user") user: User)
```

19majkel94:
And I don't like and idea of `@CurrentUser()` decorator and associated logic, I'm overwhelmed with the number of LoC which were needed to integrate this decorator and global handler function to do really simple things which can be done like this:

```typescript
export function CurrentUser() {
    return registerParamDecorator({
        required: true,
        value: actionProperties => actionProperties.request.user,
    });
}
```
`currentUserChecker` handler would still be like `actionProperties => actionProperties.request.user,` as nobody is reinventing the wheel and do JWT/Sessions logic on their own, just use existing `express-session` or `express-jwt` middlewares from npm.

19majkel94:
>updated readme and changelogs, I feel I miss something in changelogs...

You have good commit messages but yes, you've missed `removed @EmptyResultCode decorator` 😉

pleerock:
@idchlife it will throw UnauthorizedError

@19majkel94 

> I would love to see the parametrized version of @Req and @Res decorator, which should work like a @Session("property") decorator.

goal of routing-controllers is to avoid using req and res as much as possible :)

I would like `@CurrentUser` to stay. Thats very convenient feature in my opinion.

> I'm overwhelmed with the number of LoC which were needed to integrate this decorator

if problem is in number of loc, don't worry... We'll need to refactor routing-controllers codebase. I already made a small refactoring, but it is not enough. Im not satisfied with routing-controllers codebase. It lacks of good abstraction and slightly complicated. We'll do that in the future.

pleerock:
I released 0.7.0.alpha in `@next`. You can check that. Once I merge this I'll return back to typestack framework.

19majkel94:
>goal of routing-controllers is to avoid using req and res as much as possible :)

Agree, but express has many properties or getter methods on `request` that it's impossible to define dedicated decorator for each one. Also, there are many use-case when returning json is not enough, like `sendFile` which I've showed. So maybe in case of `sendFile` could I introduce the `SendFile` method decorator which would work like `@Render` - will send file with path returned from action method?

marshall007:
> Agree, but express has so many properties or getter methods on `request` that it's impossible to define dedicated decorator for each one.

@19majkel94 you could just write a custom middleware function that merges all the `request` properties you care about into `req.params`. As long as that executes before your controller method, you should be able to inject them all using the single `@Param()` decorator.

pleerock:
I don't like that people store in "req" whatever they want. First of all req is a type of `Request` and if you need to store something in there you do: `(req as any).someVar = x;` or some other ugly way. Second, to read it you again need to do that.

`@Req("property")` we can do that if people need this functionality.

regarding to `@SendFile` - there are probably other cases when people not only sending files, but other content too, including streaming. What `@SendFile` should do?

19majkel94:
> you could just write a custom middleware function that merges all the request properties you care about into req.params (...) you should be able to inject them all using the single @Param() decorator.

Wait... what?! From express doc about `req.params`:
> This property is an object containing properties mapped to the named route “parameters”. For example, if you have the route /user/:name, then the “name” property is available as req.params.name. 

So I don't think it's a good idea. Sure, you can do this, but what's the purpose? Why this:
```typescript
action(@Params() params: any) {
    const userId = params.session.user.id;
}
```
is better than simply:
```
action(@Req() { hostname }: Request)
```
Can you tell me @marshall007 ? 😉 

> I don't like that people store in "req" whatever they want. 

I haven't talked about this case. I use in my project the [`req.range()` method](https://expressjs.com/en/api.html#req.range) and I have to inject whole request object to my action to use it to partial stream file. It would be nice to be able to use `@Req("range")` decorator and inject only that method, not the entire jungle 😜 It can't be used as a `@Range(size: number)` decorator because `size` param e.g. size of file, has to be determined inside action method.

> regarding to @SendFile - there are probably other cases when people not only sending files, but other content too, including streaming. What @SendFile should do?

It should work the same way as the `@Render` decorator works:
```typescript
@Get("/files/:id")
@SendFile(options)
getFile(@Param("id") id: number) {
    return getFilePathById(id);
}
```
So just calling `response.sendFile()` method with the value returned from action method. Because now it's hard to achieve it in routing-controllers, as `res.sendFile` is asynchronous with callback, so if you don't wrap it in a promise, you will get `Can't set headers after they are sent` error 😉 
```typescript
return new Promise<void>((resolve, reject) => {
    res.sendFile(filePath), (err: any) => err ? resolve() : reject(err));
});
```

marshall007:
> Why this: [...] is better than simply:
```ts
action(@Req() { hostname }: Request)
```

@19majkel94 the reason it's better to be explicit with your `Param` and friends decorators is because you're defining the structure of your API. In theory this information could be used to create REST API documentation and/or intelligent clients by reflecting over the metadata emitted by the TypeScript compiler.

For example, if the reflection metadata tells you that `path` is a route param and `search` is a query param, you could write a client that takes an object like `{ path, search }` and intelligently builds the appropriate `GET /:path?search=:search` request.

I'm planning to work on tooling for extracting this metadata from `routing-controllers` into JSON so it can be consumed by doc generators or a client. So while there aren't super compelling reasons to decorate params properly right now, hopefully there will be soon! First step is looking into #122.

19majkel94:
> For example, if the reflection metadata tells you that path is a route param and search is a query param, you could write a client that takes an object like { path, search } and intelligently builds the appropriate GET /:path?search=:search request.

Ok but it's not the case - `hostname` is not the API but the inside implementation, as the `req.app` property is or `req.user` is. But some advanced users need this to take an advantage of full powers of express.js. The same goes to all `response` methods like `sendFile()`, `attachment()`, etc. Have I map them also to `req.params` as they are a part of REST API for swagger generation?

pleerock:
> It would be nice to be able to use @Req("range") decorator and inject only that method, not the entire jungle 

but what if range will use `this` inside, it won't work right? Overall `@Req("range")` looks dubious way of doing things, as for me its better to use `request.range()` - more clear, natural, safe and documented way of doing things

> @SendFile and @Download 

I think these decorators deserve separate issue and PR for them. Also better name for `@SendFile`. Looks really ugly... 

> Shall I map them also to req.params

no, you should not. Just use `request` as it is, whats is the problem with it. Its not a jungle, its an object with good things inside it, if you need it you should use it. You shouldn't use it only if you need stuff like query params, params etc because there is an elegant way of injecting those params. Injecting part of functions of request object isn't elegant.

19majkel94:
> but what if range will use `this` inside, it won't work right?

Just use `bind` method 😉 
```typescript
export function extractReqProp<K extends keyof Request>(propName: K, req: Request): Request[K] {
    const prop = req[propName] as object|Function;
    if (typeof prop === "function") {
        return prop.bind(req);
    } else {
        return prop;
    }
}
```
> or me its better to use request.range() - more clear, natural, safe and documented way of doing things

Yeah, it's more clear but as you said:
> goal of routing-controllers is to avoid using req and res as much as possible :)

Also @marshall007 said that the goal is:
> to create REST API documentation and/or intelligent clients by reflecting over the metadata emitted by the TypeScript compiler.

So `range` is a part of API - it's a parsed `Content-Range` header value used to serve partial content. I know that's not a really common case in typical REST API, but it's used in streaming services.

> I think these decorators deserve separate issue and PR for them. Also better name for @SendFile. Looks really ugly...

It's just a name of Response method. More intuitive for express user than a fancy name not matching the method. But the best way would be to introduce custom method decorators which could work just like `interceptor`. We could then receive returned value from metod and do `res.sendFile(returnedValue)`.

pleerock:
> goal of routing-controllers is to avoid using req and res as much as possible :)

everything is relative, right?) avoid, but not for such a dirty operations.

marshall007:
> So `range` is a part of API - it's a parsed `Content-Range` header value used to serve partial content.

In that case you would want to use the `@HeaderParam` decorator directly, that way you could inform the doc generator/other tooling that this endpoint expects the header. I'd then use a transform (or `class-validator` if #122 lands) to parse the header value before it's injected into your controller method.

19majkel94:
> to parse the header value before it's injected into your controller method

It can't be fully parsed AOT, it needs `the maximum size of the resource` param which is determined inside action param (find filename correlated with entity with id from param and read file stat from disk).

> you would want to use the @HeaderParam decorator directly

In term of self-described API - yes, but the goal of routing-controllers and DI is to provide an method API which focus on the problem (return entity by id) not the way to achieve it (get a param, parse it to get value). It's hard to test if you have to create a http header string instead of pass simple { start, end } object`. 

The same problem goes to JWT or Session with your solution - I should use `@HeaderParam("Authentication")` and then inside action method parse it by hand to get the current user object? It's really not developer friendly, your tool should recognize `@JWT` decorator and generate header description in API definition. I just receive ready for use `User` object so I don't have to generate JWT token string and create header string in case of unit testing. And since param decorators are for injection, I can just change in my code `@Session("user")` to `@JWT()` and the whole logic and testing stays the same, no need to change in case of authentication way.

I have an idea of `@Range` decorator which should be nice for API documentation generation tool and also developer-friendly in use. The sample usage would be like this:
```typescript
action(@Range(options) reqRange) {
    const maxSize = getFromSomewhere();
    const range = reqRange(maxSize);
    // do something with array of parsed range from headers
}
```
where `options` could be e.g. `{ accept: "bytes" }` or sth like that.

PR #136, opened 2017-05-11
Changed @types dependencies to be under dev-dependencies

vlio20:
I think that it is better to put the `@types` under the `devDependencies` than the `dependencies`

### DIFF
--- a/README.md
+++ b/README.md
@@ -88,7 +88,7 @@ You can use routing-controllers with [express.js][1] or [koa.js][2].
 
     Optionally you can also install their typings:
 
-    `npm install @types/express @types/body-parser @types/multer --save`
+    `npm install @types/express @types/body-parser @types/multer --save-dev`
 
     **b. If you want to use routing-controllers with *koa 2*, then install it and all required dependencies:**
 
@@ -96,7 +96,7 @@ You can use routing-controllers with [express.js][1] or [koa.js][2].
 
     Optionally you can also install their typings:
 
-    `npm install @types/koa @types/koa-router @types/koa-bodyparser --save`
+    `npm install @types/koa @types/koa-router @types/koa-bodyparser --save-dev`
     
 4. Its important to set these options in `tsconfig.json` file of your project:
 


pleerock:
not sure if it is correct. adding @19majkel94

19majkel94:
The rules are simple:
- if you build a library which other devs will use in their projects, put `@types` in dependecies section
- if you build a project (backend app) which won't be a common module or something, you can place typings in `devDependencies` section, unless you push source code to production and where Typescript compilation is a part of build step (postinstall script) - in this case it has to be a dependency because tsc will throw error about missing `@types`

vlio20:
@19majkel94, where is the reference for this rules? 
For example, Angular puts all its types under dev dependencies - because the `@types` aren't part of the production code.

19majkel94:
>  where is the reference for this rules?

https://github.com/Microsoft/types-publisher/issues/81

>Angular puts all its types under dev dependencies - because the @types aren't part of the production code

Nope. They have e.g. `@types/fs-extra` as dev dependencies because it's a part of build tool and `fs-extra` is not listed in dependencies. It mean that when you build an angular app, you won't use their build tools to make an app.

Let's take an other example - fs-promise. It was a wrapper around node` fs` and `fs-extra` which added promise support. So it introduce some new function and re-export the ones from fs and fs-extra. So in order to use it, their `@types` has to be an dependency, because npm won't install dev dependencies of your dependencies and `tsc` throw error because you use something that is not a part of package. I have experienced this issue with this fail commit: https://github.com/kevinbeaty/fs-promise/commit/90fd3d42d4b48a88fa46c69f03825534e7c791e2

So to clarify things:
- install `@types` as dependency for dependencies
- install `@types` as dev dependency for your dev dependencies

Of course with frontend apps bundled with webpack you can place `@types` as a dev dependencies because there are not present in bundle at all but it's an exception and complicate the rules, so simply remember the rule above and everything will be working 😉

pleerock:
so... we can close this one?

pleerock:
I guess yes.

PR #148, opened 2017-05-18
Fix Session param decorator missing options

### LINKED ISSUE #140: Rationale for not supporting "required" session params?
It seems like `@Session(<param>, { required: <true|false> })` should work to be consistent with `@Param`, `@QueryParam`, `@Body`, etc.

I can maybe understand why `@Session({ required: false })` isn't supported, but even so I think it's unnecessarily opinionated to not support the `required` option on session params.

19majkel94:
Fixes #140.

### DIFF
--- a/src/decorator/Session.ts
+++ b/src/decorator/Session.ts
@@ -1,20 +1,39 @@
-import {getMetadataArgsStorage} from "../index";
+import { getMetadataArgsStorage } from "../index";
+import { ParamOptions } from "../decorator-options/ParamOptions";
+
 
 /**
  * Injects a Session object to the controller action parameter.
  * Must be applied on a controller action parameter.
  */
-export function Session(objectName?: string): Function {
+export function Session(options?: ParamOptions): ParameterDecorator;
+/**
+ * Injects a Session object to the controller action parameter.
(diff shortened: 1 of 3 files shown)


src/decorator/Session.ts
```diff
@@ -1,20 +1,39 @@
...
+export function Session(propertyName: string, options?: ParamOptions): ParameterDecorator;
+
+export function Session(optionsOrObjectName?: ParamOptions|string, paramOptions?: ParamOptions): ParameterDecorator {
+    let propertyName: string|undefined;
+    let options: ParamOptions|undefined;
+    if (typeof optionsOrObjectName === "string") {
+        propertyName = optionsOrObjectName;
+        options = paramOptions || {};
+    } else {
+        options = optionsOrObjectName || {};
```
marshall007: @19majkel94 based on the conversation in #140 I thought we decided it didn't make sense to support `@Session({ required: false})`, which I think this would allow you to do now.

19majkel94: Yeah, I know but I think that default `true` is enough. If someone want to hurt himself, it will allow to do it 😉

PR #166, opened 2017-05-26
dynamic redirect

Diluka:
1. empty result behave like old one
2. string result override the default in @Redirect
3. object result using template-url to generate url

https://github.com/pleerock/routing-controllers/issues/159

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,5 @@ build/
 node_modules/
 coverage/
 npm-debug.log
+.vscode/
+.idea/
--- a/README.md
+++ b/README.md
@@ -540,6 +540,29 @@ getUsers() {
 }
 ```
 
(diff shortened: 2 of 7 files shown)


marshall007:
@Diluka I think it would also make sense to support `@Redirect()` with no arguments and allowing the controller method to return a string URL.

Diluka:
@marshall007 
Diving into the driver, it decide whether there have redirections by checking the redirect string if not empty.

pleerock:
You forgot to make url argument optional in `@Redirect` decorator. Also please add docs regarding to this new functionality.

Diluka:
@pleerock 
url argument is designed to be not optional and adding docs

pleerock:
Thank you! Released in `0.7.0-alpha.15`

PR #177, opened 2017-06-07
feat(Controller): add date type handling to action parameters

Efk3:
First of all: thank you for the awesome library.

I recently started to use it and see that there is no solution (or I simple didn't find it) for receive Date type in a query param so I made a little modification to make that possible.

### DIFF
--- a/src/ActionParameterHandler.ts
+++ b/src/ActionParameterHandler.ts
@@ -125,6 +125,13 @@ export class ActionParameterHandler {
 
                 return !!value;
 
+            case "date":
+                const parsedDate = new Date(value);
+                if (isNaN(parsedDate.getTime())) {
+                    return Promise.reject(new BadRequestError(`${param.name} is invalid! It can't be parsed to date.`));
+                }
+                return parsedDate;
+
             default:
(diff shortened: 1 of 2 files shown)


src/ActionParameterHandler.ts
```diff
@@ -125,6 +125,13 @@ export class ActionParameterHandler {
 
                 return !!value;
 
+            case "date":
+                const parsedDate = new Date(value);
+                if (isNaN(parsedDate.getTime())) {
+                    throw new BadRequestError(`${param.name} is invalid! It can be parsed to date.`);
```
pleerock: I didn't like new try/catch block in next file. Can you try here do `Promise.reject(new BadRequestError(`${param.name} is invalid! It can be parsed to date.`))`; instead of ` throw new BadRequestError(`${param.name} is invalid! It can be parsed to date.`);` ? I think it should work without having to add extra try catch block

Efk3:
I fixed it. It works with Promise.reject as you suggested.

pleerock:
Thank you! It will be released in a next `0.7.1` version

PR #178, opened 2017-06-08
Fix/auth decorator status code

Diluka:
Let `@Authorized()` raising `401` and `@Authorized("some_role")` raising `403`.

### DIFF
--- a/src/driver/express/ExpressDriver.ts
+++ b/src/driver/express/ExpressDriver.ts
@@ -12,6 +12,7 @@ import {AccessDeniedError} from "../../error/AccessDeniedError";
 import {AuthorizationCheckerNotDefinedError} from "../../error/AuthorizationCheckerNotDefinedError";
 import {isPromiseLike} from "../../util/isPromiseLike";
 import {getFromContainer} from "../../container";
+import {AuthorizationRequiredError} from "../../error/AuthorizationRequiredError";
 const cookie = require("cookie");
 const templateUrl = require("template-url");
 
@@ -117,7 +118,9 @@ export class ExpressDriver extends BaseDriver implements Driver {
                 if (isPromiseLike(checkResult)) {
                     checkResult.then(result => {
                         if (!result) {
-                            return this.handleError(new AccessDeniedError(action), actionMetadata, action);
+                            return this.handleError((actionMetadata.authorizedRoles.length === 0
+                                    ? new AuthorizationRequiredError(action) : new AccessDeniedError(action)),
+                                actionMetadata, action);
 
                         } else {
                             next();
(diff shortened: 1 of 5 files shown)


src/driver/express/ExpressDriver.ts
```diff
@@ -117,15 +118,19 @@ export class ExpressDriver extends BaseDriver implements Driver {
...
+                                actionMetadata, action);
 
                         } else {
                             next();
                         }
                     });
                 } else {
                     if (!checkResult) {
-                        this.handleError(new AccessDeniedError(action), actionMetadata, action);
+                        this.handleError((actionMetadata.authorizedRoles.length === 0
```
pleerock: all these code blocks become messy... maybe we should avoid using ternary if and beatify code?

19majkel94: ternary is faster, so if the libs want's to be fast, this should stay with other tail call optimizations 😜

NoNameProvided: > so if the libs want's to be fast

I see you need time to heal from the traumatic experience of that conversation 😄

19majkel94: The only traumatic thing is the hypocrisy:
- `Promise.resolve` removes half on duplicated code block but it's slower, so we stick with messy code because "its important to make it performant as we can"
- now "all these code blocks become messy", so we should remove ternary and tail call optimization to "beatify code"

So please be consistent - peformance at any cost or readable and maintainable code 😄

src/driver/express/ExpressDriver.ts
```diff
@@ -114,21 +115,24 @@ export class ExpressDriver extends BaseDriver implements Driver {
...
-                            return this.handleError(new AccessDeniedError(action), actionMetadata, action);
 
-                        } else {
-                            next();
-                        }
-                    });
-                } else {
-                    if (!checkResult) {
-                        this.handleError(new AccessDeniedError(action), actionMetadata, action);
+                const handleError = (result: any) => {
```
Diluka: So, do you like this one?

src/driver/express/ExpressDriver.ts
pleerock: @19majkel94 you should investigate a bit how `Promise.resolve` actually works, its not as simple as let vs const. We make decisions based on pros and cons of everything, its impossible to be consistent in everything, so make judging based on each scenario individually.

src/driver/express/ExpressDriver.ts
```diff
@@ -114,21 +115,24 @@ export class ExpressDriver extends BaseDriver implements Driver {
...
+                    if (!result) {
+                        return this.handleError((
```
pleerock: multiline params this way do not look good, simplify this. extract error class in a separate variable if needed

pleerock:
everyone thinks its good to merge (not only performance and code style considerations 😆 ) ?

PR #189, opened 2017-06-22
Fix instructions about creating interceptors

netes:
Hi there,
I've noticed a small bug in the interceptor documentation. There is uneeded decorator `@Interceptor` in the section: https://github.com/pleerock/routing-controllers#interceptor-classes. It makes it global which is not desired behaviour.

PS. Sorry for whitespaces "mess" - my editor automatically changed that. Here's the actual change:
https://github.com/pleerock/routing-controllers/pull/189/files#diff-04c6e90faac2675aa89e2176d2eec7d8L1033


pleerock:
Thank you!

Issue #204, opened 2017-07-09
Proposal: response wrappers for success responses


adnan-kamili:
In case of errors we can throw exceptions like NotFoundError etc. Similarly for success responses their could be wrappers which do common tasks like:

taken from Asp.Net Core:

```
// return response with 200 status code - current default
return Ok(user);  
```

```
// return response with 201 status code, add url (first param) to Location header
return Created($"/v1/users/{user.Id}", new { message = "User was created successfully!" });
```

```
// return empty response with 204 status code
return NoContent();
```

Using decorators like `@HttpCode(204)` for every overriding behaviour is sometimes overkill, it adds such a big stack of decorators for each controller action.

Currently to add location header for 201 response, one would need to create a middleware, but it's very common in rest api's should be handled at framework level.

NoNameProvided:
You can use [interceptors](https://github.com/pleerock/routing-controllers#using-interceptors).

adnan-kamili:
It's not about formatting response body, it's about support for REST compliant response status/headers out of the box, purpose of framework/library is to care for this common stuff.

19majkel94:
`routing-conrollers` is built on top of `express` framework. The main idea is to provide an abstraction over express and basic http-related things, so the action handler is the normal method like a service, which takes User object, id number and return entity from database. Injecting the params from request fields and returning response is happening under the hood, so you can easily create unit tests.

If you need some non standard things, like pipe filestream to response, you can use express features. Your 201 example:
```ts
response
  .status(201)
  .location(`/v1/users/${user.id}`)
  .send({ message: "User was created successfully!" });
```
And thanks to interceptor functionality and `routing-controllers` extensibility, you can create your own wrapper classes like `Created` or `NoContent` and transform it there to correct responses 😉

pleerock:
@adnan-kamili what is your proposal here? As I understood you want to introduce functions like `Ok`, `Created`, `NoContent`? If yes, then answer is that its a bad idea. Because first we have decorators for same purpose which are better then function but less flexible, and second we have standard express API you can use to do what you want, e.g.

```ts
action(@Res() res: Response) {
    res.status = 201;
    res.location = ... etc
}

**Comment by adnan-kamili** [comment]
I am using koa and it actually breaks the symmetry of the code. It requires the explicit setting of headers:
```javascript
ctx.set("Location", `/v1/users/${user.id}`);
return { message: "user created successfully!" };
```

[...]

PR #209, opened 2017-07-11
allow default options

chaowlert:
Hi @pleerock,

Your library is great. And I think it would be better if we are able to set default values.

For example, if method is `void` or `Promise<void>`, I expected it to return `204` rather than `404` by default. If there is any exception, then I will override by `OnUndefined` decorator.

The other setting I prefer to set is `required` option, I want all parameters to required by default. And I can override with the param options.

This would give more flexibility to other projects which have different convention, and improve readability since there is less number of settings needed to be specified.

Thank you,

### DIFF
--- a/src/RoutingControllersOptions.ts
+++ b/src/RoutingControllersOptions.ts
@@ -91,12 +91,12 @@ export interface RoutingControllersOptions {
      */
     defaults?: {
         /**
-         * If set, all null response will return specified status code by default
+         * If set, all null responses will return specified status code by default
          */
         nullResultCode?: number;
 
         /**
-         * If set, all undefined response will return specified status code by default
+         * If set, all undefined responses will return specified status code by default
          */
         undefinedResultCode?: number;
 
--- a/src/RoutingControllersOptions.ts
+++ b/src/RoutingControllersOptions.ts
@@ -1,7 +1,8 @@
-import {ClassTransformOptions} from "class-transformer";
-import {ValidatorOptions} from "class-validator";
 import {AuthorizationChecker} from "./AuthorizationChecker";
+import {ClassTransformOptions} from "class-transformer";
 import {CurrentUserChecker} from "./CurrentUserChecker";
+import { ParamOptions } from "./decorator-options/ParamOptions";
+import {ValidatorOptions} from "class-validator";
 
 /**
  * Routing controller initialization options.
@@ -85,4 +86,28 @@ export interface RoutingControllersOptions {
      */
     currentUserChecker?: CurrentUserChecker;
     
+    /**
+     * Default settings
+     */
+    defaults?: {
+        /**
+         * If set, all null response will return specified status code by default
+         */
+        nullResultCode?: number;
+
+        /**
+         * If set, all undefined response will return specified status code by default
+         */
+        undefinedResultCode?: number;
+
+        /**
+         * Default param options
+         */
+        paramOptions?: {
+            /**
+             * If true, all non-set parameters will be required by default
+             */
+            required?: boolean;
+        };
+    };
 }
--- a/README.md
+++ b/README.md
@@ -41,6 +41,7 @@ You can use routing-controllers with [express.js][1] or [koa.js][2].
       - [Render templates](#render-templates)
       - [Throw HTTP errors](#throw-http-errors)
       - [Enable CORS](#enable-cors)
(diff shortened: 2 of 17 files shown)


NoNameProvided:
> For example, if method is void or Promise<void>, I expected it to return 204

if you return null it will create a 204 response

chaowlert:
@NoNameProvided I know that if we return null, we will get 204. But some other projects might would like to return 404 to represent data not found.

pleerock:
adding @19majkel94 opinion on this PR

19majkel94:
> return 404 to represent data not found

In JS world, if something is not present, it's represented by `undefined` which will create a 404 response in this case.
Null is a consciously empty value, like in TypeORM when we want to delete property/relation, and here in `routing-controllers` to indicate that we consciously want to return 204 response.

But the idea of default options for decorators is great, for me required should be on by default but it's a breaking change and not everyone likes it.

For explicit setting up response 204/404 rather than null/undefined maybe we should implement #204?

src/RoutingControllersOptions.ts
```diff
@@ -85,4 +86,28 @@ export interface RoutingControllersOptions {
...
+
+        /**
+         * If set, all undefined responses will return specified status code by default
+         */
+        undefinedResultCode?: number;
+
+        /**
+         * Default param options
+         */
+        paramOptions?: {
```
pleerock: no need to introduce a new object, simply add `paramsAlwaysRequired` option

19majkel94: But in the future we might need to introduce new param options (transform/validate?), so I think it's better if it's extendable so we wouldn't have to introduce breaking change.
 `paramsAlwaysRequired: true,` isn't much shorter than `paramOptions: { required: true },` 😉

chaowlert:
Is there anyway to run unittest without node 7? One of legacy project not support node 7.

19majkel94:
@chaowlert 
Use https://github.com/creationix/nvm or https://github.com/coreybutler/nvm-windows.

chaowlert:
@19majkel94 I added unit test and example per your comment

19majkel94:
@chaowlert Thanks but looks like the feature works both with express and koa. For testing this we use `assertRequest`. The code of sample test case looks like this:
```ts
let expressApp: any, koaApp: any;
before(done => expressApp = createExpressServer({ classTransformer: true }).listen(3001, done));
after(done => expressApp.close(done));
before(done => koaApp = createKoaServer({ classTransformer: true }).listen(3002, done));
after(done => koaApp.close(done));

assertRequest([3001, 3002], "post", "users", { firstName: "Umed", lastName: "Khudoiberdiev" }, response => {
    expect(initializedUser).to.be.instanceOf(User);
    expect(response).to.have.status(200);
});
```
So please refactor your test case code to follow our used pattern and change `describe` text to `"routing-controllers defaults options"` or sth like that. I would be able then to merge this PR 😉

chaowlert:
@19majkel94 done

src/RoutingControllersOptions.ts
pleerock: we already have options for transform/validate. `paramsAlwaysRequired` is less complexier then `paramOptions: { required: true }`. Its pointless to have object for options when there is only one option. If we'll have something in the future we will make deprecations if needed - its easy to do that in here.

NoNameProvided: I agree with @19majkel94 it's more future proof to add it as an object, it would save as a few release where we would have to keep the deprecated version working as well. 

Btw is this PR ok except this part?

19majkel94: For me it's ok. Waiting for green light 🚥 from @pleerock

pleerock: Okay lets left it as it is if you both think its better. But Im still against such redundant doubtful-value nestings which only brings complexity.

twittwer:
Really helpful feature :+1: 
When will it be available via npm?

IAMtheIAM:
+1 please release this feature on npm :-)

19majkel94:
>  pleerock published 15 hours ago
0.7.2 is the latest of 55 releases

https://www.npmjs.com/package/routing-controllers

IAMtheIAM:
Awesome! perfect timing :-)

Issue #220, opened 2017-07-18
One url make 2 action call


soulski:
Hi, I have one Controller which have 2 method like below

```javascript
import { JsonController, Get } from 'routing-controllers'

@JsonController('/api')
export class APIController {

    @Get('/user')
    public getByUser() {
        console.log('output2')
        return Promise.resolve({ say: 'user' })
    }

    @Get('/:id')
    public getById() {
        console.log('output1')
        return Promise.resolve({ say: 'hello1' })
    }

}
```

I run server and then send http request in path '/user'. The result in console log print follow below

>output1 
>output2

and I receive response as json { say: 'user' }. I take a look at source code and found that after action was called, routing-controller call next method which I suspect it is the reason why 2 action which match path url call. I test by insert after after middleware that didn't call next. The result is only getByUser was called. Is this intend behavior? It use invoke only getByUser, is't it?

NoNameProvided:
See https://github.com/pleerock/routing-controllers/issues/213#issuecomment-315733755 for details

marshall007:
@NoNameProvided this still seems like a bug though. It shouldn't be possible for controller methods to fall through like middleware. Once a controller method has handled the request that should be the end of it.

NoNameProvided:
Yeah, it's not mentioned in the docs and this is what makes it seems like a bug the most, also that you can only return response from the first route which was hit. PR-s are welcomed for this as for any other issue, but first lets ask @pleerock if it was intentional or not?

briman0094:
I'm running into this issue as well. I have an action which listens on `/js/common.js` and one which listens on `/js/:path*`, with `/js/common.js` being first. In Express, if a route were to hit `/js/common.js`, it would immediately stop propagation and run that action. With `routing-controllers`, it's hitting `/js/common.js` first and then falling through to `/js/:path*`, causing an error when setting some headers, because they've already been sent by `/js/common.js`. Routes should not fall through.

briman0094:
I just made a pull request with a solution to this problem that I have now been using in my application (see #222). Place `@Terminates()` on your most specific route (and make sure to define it first).

If you don't feel like waiting for the PR to be accepted (assuming it is), clone my `issue-220` branch, run `npm install && gulp package`, and replace the contents of your `node_modules/routing-controllers` folder with the new contents of `build/package` in the cloned repo (tedious, but it works).

NoNameProvided:
[...]

PR #222, opened 2017-07-19
Add @Terminates() decorator to prevent fallthrough and automatic response handling

briman0094:
Placing `@Terminates()` on an action prevents it from falling through to any future actions or middleware (except, of course, for error handling middleware in the event of an error). This is useful for routes which manually send reponse data or set headers, or for routes which call external handlers such as Browserify (passing `req` and `res` and letting the external handler provide content).

This decorator also prevents overlapping routes from conflicting. For example, if there are two actions defined, one for `/js/bundle.js` and one for `/js/:file(*)`, under the master branch, both actions get called (see #220) which can cause issues with headers, etc. When `@Terminates()` is placed on `/js/bundle.js`, it no longer falls through to `/js/:file(*)` and properly terminates the route chain.


briman0094:
I'm closing this in favor of explicit fallthrough instead of explicit termination. After more troubles with my app, it makes sense to require a decorator to fall through routes instead of assuming all routes should fall through. It's very uncommon for a route handler in Express to actually have a `next` parameter unless it is calling it for error handling purposes, which is already handled in `routing-controllers`. `next` shouldn't be called after an action handler runs unless the user opts in to it, in my opinion, simply because that's not how Express apps normally work.

PR #223, opened 2017-07-19
Add @Fallthrough() decorator and global option to enable route fallthrough, which is now off by default

briman0094:
Route fallthrough (route handlers calling `next()` without parameters) is now off by default, as it is not the implicit behavior in Express (I don't use Koa, but it appears to have similar paradigms to Express). If a route needs to fall through to other routes or "after" middleware, `@Fallthrough()` can be placed on the route handler. If the user wants all routes to fallthrough by default, the new `automaticFallthrough` global option can be enabled when initializing `routing-controllers`.

### DIFF
--- a/src/RoutingControllersOptions.ts
+++ b/src/RoutingControllersOptions.ts
@@ -20,6 +20,11 @@ export interface RoutingControllersOptions {
     routePrefix?: string;
 
     /**
+     * Enable automatic fallthrough from action handlers to future actions or middleware
+     */
+    automaticFallthrough?: boolean;
+
+    /**
      * List of controllers to register in the framework or directories from where to import all your controllers.
      */
     controllers?: Function[]|string[];
--- a/src/decorator/Fallthrough.ts
+++ b/src/decorator/Fallthrough.ts
@@ -1,0 +1,15 @@
+import {getMetadataArgsStorage} from "../index";
+
+/**
+ * Indicates that this route should fall through to further route handlers or middleware after being executed.
+ * This is equivalent to a route handler calling next() with no parameters in Express.
+ */
+export function Fallthrough(): Function {
+    return function (object: Object, methodName: string) {
+        getMetadataArgsStorage().responseHandlers.push({
+            type: "fallthrough",
+            target: object.constructor,
+            method: methodName
+        });
+    };
+}
--- a/src/driver/BaseDriver.ts
+++ b/src/driver/BaseDriver.ts
@@ -56,6 +56,11 @@ export class BaseDriver {
     routePrefix: string = "";
 
     /**
+     * Enable automatic fallthrough from action handlers to future actions or middleware
+     */
+    automaticFallthrough: boolean;
+
+    /**
      * Indicates if cors are enabled.
      * This requires installation of additional module (cors for express and kcors for koa).
      */
--- a/src/driver/Driver.ts
+++ b/src/driver/Driver.ts
@@ -60,6 +60,11 @@ export interface Driver {
     routePrefix: string;
 
     /**
+     * Enable automatic fallthrough from action handlers to future actions or middleware
+     */
+    automaticFallthrough: boolean;
+
+    /**
      * Indicates if cors are enabled.
      * This requires installation of additional module (cors for express and kcors for koa).
      */
--- a/src/driver/express/ExpressDriver.ts
+++ b/src/driver/express/ExpressDriver.ts
@@ -258,7 +258,8 @@ export class ExpressDriver extends BaseDriver implements Driver {
                 options.response.redirect(action.redirect);
             }
 
-            options.next();
+            if (this.automaticFallthrough || action.fallthrough)
+                options.next();
 
         } else if (action.renderedTemplate) { // if template is set then render it
             const renderOptions = result && result instanceof Object ? result : {};
@@ -273,7 +274,9 @@ export class ExpressDriver extends BaseDriver implements Driver {
                 } else if (html) {
                     options.response.send(html);
                 }
-                options.next();
+
+                if (this.automaticFallthrough || action.fallthrough)
+                    options.next();
             });
 
         } else if (result !== undefined || action.undefinedResultCode) { // send regular result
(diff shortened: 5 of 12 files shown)


NoNameProvided:
Hi @briman0094! 

Thanks for the PR! The test (an so the implementation) will need some modification. In routing-controllers every _after middleware_ should be executed when a route had been hit. The `@Fallthrough()` decorator should only prevent calling the following matching routes but not the _after middlewares_.

@19majkel94 @pleerock thoughts? I like this, but when implemented this will be a breaking change, however I doubt anyone took advantage of the route fallthrough.

marshall007:
I'm not even really convinced that the `automaticFallthrough` config option and `@Fallthrough()` decorator should be exposed. I think the framework should take a more opinionated approach and disallow this behavior altogether. Given that we already have before/after middleware with plenty of configurability, I don't see a valid use case for allowing controller methods to fall through.

If we expose this stuff as part of the documented public API I think we're just encouraging people to do bad things.

src/RoutingControllersOptions.ts
```diff
@@ -20,6 +20,11 @@ export interface RoutingControllersOptions {
     routePrefix?: string;
 
     /**
+     * Enable automatic fallthrough from action handlers to future actions or middleware
+     */
+    automaticFallthrough?: boolean;
```
pleerock: `fallthrough` is better then `automaticFallthrough `. Also TBH `fallthrough` does not indicate for me what this option does, but I can't find a good name for it. Maybe this terminology is highly used, simply Im not familiar with it, adding other thoughts @19majkel94 @NoNameProvided  . Is this word better then something like `callNext: boolean` ?

pleerock:
So here we faced again with problem that framework calls `next` after.

I don't like decorator this PR provides, it seems to me that one global option is enough. 

I'm okay to disable `next` call after each controller action by default.
What are you thoughts on it?

@19majkel94 can you please help me to remember what issues it will cause? I think "after" middlewares won't work, am I right?

src/RoutingControllersOptions.ts
NoNameProvided: > Also TBH fallthrough does not indicate for me what this option does, but I can't find a good name for it. Maybe this terminology is highly used, simply Im not familiar with it

It should be `FallThrough` because there is no such word as _fallthrough_, it is _fall through_. It is used in programming so I am cool with it, however we can make it more explicit via naming it as `routeFallThrough`.

NoNameProvided:
> @19majkel94 can you please help me to remember what issues it will cause? I think "after" middlewares won't work, am I right?

I think this is the related discussion: https://github.com/pleerock/routing-controllers/issues/114#issuecomment-296802216 and here https://github.com/pleerock/routing-controllers/issues/50. You are right the problem was that no after middlewares are called.

The solution would be to handle middlewares at routing-controller level instead of express, right?

briman0094:
I thinking handling after middle are at the routing level is the only solution. Express doesn't really have a concept of "after middleware" because it sends headers after a matched route handler is called. If we want anything to be called after a `routing-controllers` method, and before headers are sent, I think it will have to be in either `callAction` or `handleSuccess` of the driver.

briman0094:
Thanks for the feedback...I'll work on this a bit later today and into the next week.

soulski:
Any progress on this PR?

19majkel94:
ping @briman0094

briman0094:
My apologies...I've been swamped at work. I'll make a note to make some progress on this over the weekend...hopefully I can finish it by the end of the weekend.

soulski:
Hi @briman0094, I understand that you're really busy and I really need this PR so would you mine if I copy all of your code and send another PR or I can send PR to your repository to make this PR go through.

briman0094:
Yeah, that's fine. I feel bad for having to keep delaying the PR; I just haven't had the time to look at it yet with all my other projects. You should just be able to fork my fork of this repo, check out the branch for this PR, make your changes, and then make a new PR from your branch without having to copy/paste any code.

19majkel94:
@soulski Contributions and PRs are welcome but please take into account [my comments](https://github.com/pleerock/routing-controllers/pull/223#pullrequestreview-54538428) 😉

briman0094:
Thanks for taking charge of this; it's really a load off my shoulders 😄

soulski:
@19majkel94 After had a look on code, I found that If we don't call `next`, It's not only break after middleware but it also break before middleware in case Koa. Look like this will not be easy fix as I thought because we need to handle different flow on both Express and Koa. Do you have any suggestion?

19majkel94:
> it also break before middleware in case Koa

But why? All koa's before midlewares do `await next()` so when we won't call next in our handler, it just return control down in the call stack to the first middleware.

After middlewares are really weird thing in routing-controllers, not used in koa (there is `await next()` for calling the after handler complete) and in express (everybody patch the `.end()` method).
You just need to get the after middlewares array in metadata storage and manually call it like in koa it would be `return nextMiddleware(ctx, next)`.

soulski:
I am mis-understanding koa-router. I just found that if there is middleware that didn't call next append as last middleware, router also match pattern won't be call. So what do you think if I found that routeFallThrough is off, I will put `Terminal middleware` which is middleware that doesn't call next in the last queue of middleware. This will help me doesn't need to simulate call after middleware.

19majkel94:
Bare koa handler doesn't have to call next to make middlewares work. Here is a sample from koa's website:
```js
const Koa = require('koa');
const app = new Koa();

// x-response-time

app.use(async (ctx, next) => {
  const start = Date.now();
  await next();
  const ms = Date.now() - start;
  ctx.set('X-Response-Time', `${ms}ms`);
});

// logger

app.use(async (ctx, next) => {
  const start = Date.now();
  await next();
  const ms = Date.now() - start;
  console.log(`${ctx.method} ${ctx.url} - ${ms}`);
});

// response

app.use(async ctx => {
  ctx.body = 'Hello World';
});

app.listen(3000);
```

jschutijzer:
When is this going to be released?

PR #225, opened 2017-07-19
Use response.render instead of express.render to provide locals correctly

briman0094:
`res.render` automatically provides `res.locals` to the view engine, whereas `app.render` does not. Use `res.render` so that `res.locals` properties work correctly in `@Render` actions.

### DIFF
--- a/src/driver/express/ExpressDriver.ts
+++ b/src/driver/express/ExpressDriver.ts
@@ -263,7 +263,7 @@ export class ExpressDriver extends BaseDriver implements Driver {
         } else if (action.renderedTemplate) { // if template is set then render it
             const renderOptions = result && result instanceof Object ? result : {};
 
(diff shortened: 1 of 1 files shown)


19majkel94:
Do we have unit tests for `@Render` or we need tests for this PR (rendering with locals)?

NoNameProvided:
> we need tests for this PR (rendering with locals)

Yeah, thats a good idea.

NoNameProvided:
@briman0094 Can you add a test what checks that locals are used for rendering the template?

19majkel94:
Ping @briman0094 - could you make a simple test case which test the new feature (locals are used for rendering the template)? You should extend [this file](https://github.com/pleerock/routing-controllers/blob/master/test/functional/render-decorator.spec.ts) with a few lines 😉

NoNameProvided:
Closing as no reply from the author, has been re-implemented in #274

PR #233, opened 2017-08-01
handle custom errors thrown by authorization checker

twittwer:
Really great module 👍 

One requirement in our project is to differentiate Unauthorized and Forbidden code during authorization process.
That's why I added an catch to the authorizationChecker, to be able to throw HttpErrors the same as in controllers.
I think that could be helpful for others too.

```typescript
createExpressServer({
    authorizationChecker: async (action: Action, roles: string[]) => {

        const token = action.request.headers["authorization"];
        if (!token)
            throw new UnauthorizedError();

        const user = await getEntityManager().findOneByToken(User, token);
        if (!user)
            throw new UnauthorizedError();

        if (!roles.length)
            return true;
        if (roles.find(role => user.roles.indexOf(role) !== -1))
            return true;

        throw new ForbiddenError();
    }
});
```

Thanks

### DIFF
--- a/src/driver/express/ExpressDriver.ts
+++ b/src/driver/express/ExpressDriver.ts
@@ -126,7 +126,9 @@ export class ExpressDriver extends BaseDriver implements Driver {
                 };
 
                 if (isPromiseLike(checkResult)) {
-                    checkResult.then(result => handleError(result));
+                    checkResult
+                        .then(result => handleError(result))
+                        .catch(error => this.handleError(error, actionMetadata, action));
                 } else {
                     handleError(checkResult);
                 }


festen:
I have the same requirement (but in Koa) as well, I imagine it is more or less the same logic/different file?. It now only throws code 500 errors. Excited about this feature 👍

19majkel94:
I don't know why build test failed on the newest node.js engine:
```
1) action parameters @Session(param) should throw required error when param is empty asserting port 3001:
     TypeError: Cannot read property 'statusCode' of undefined
```
But it's not related to this feature so I will do the honors and merge this PR with 1 new line 😉

NoNameProvided:
At first glance without digging, it seems this issue exists in the koa part as well. Why did we merge this without fixing there as well? We should aim consistency between the two driver. 

The related part is [KoaDriver.ts#L106](https://github.com/pleerock/routing-controllers/blob/master/src/driver/koa/KoaDriver.ts#L106 )

NoNameProvided:
If no one fixes this, I will do it in the evening. Anyone should feel free to create a patch for this.

19majkel94:
@NoNameProvided You're right! 😞 I've spend whole day digging through issues and PR and I haven't noticed that. It would be great if you would create a patch for koa and even create a test case for this feature 😉

NoNameProvided:
> BTW I see more and more repeated code in drivers, I think it's time to make a big refactor to keep DRY.

I agree on that. I will do it.

twittwer:
@NoNameProvided Sorry, for not answering - Just missed the ongoing conversation 🙈

NoNameProvided:
Patch created in #247.

Issue #231, opened 2017-08-01
Returning undefined in a JsonController does not return JSON.


GeeWee:
A JSONController that returns undefined, returns a 404, but with the content type set to html, and a html document with the error message.

Simple repro:

```
@JsonController('/widget')
export class WidgetController {
	@Get('/:id')
	public async getOne(@Param('id') id: number) {
		return undefined;
	}
}
```

and a test using supertest
```
const response = await request(app)
			.get('widget/1')
			.expect('Content-Type', /json/)
```

I think it would make more sense for the error to be returned in json format.

19majkel94:
The problem is here:
https://github.com/pleerock/routing-controllers/blob/master/src/driver/express/ExpressDriver.ts#L279
And in koa driver too!

`undefined` goes throught whole `if-else` tree and goes up to the last `else` statement which call `next()`. So it behaves like there were no registered route and the response shape depends on how the underlying framework handle it.

And you're right, this shouldn't work like this, I will try to fix it now 😉

GeeWee:
Wonderful!

pleerock:
since PR is merged I think this can be closed

PR #241, opened 2017-08-06
Fix 404 text content-type for JSON controllers

19majkel94:
Solving #231.

I decided to throw `NotFoundError` when `undefined` is returned from an action. This should let error handler decide about formatting the JSON response for JSON controllers and text/plain or html for text responses.

However I think we should place the info about the route where the undefined has been returned. @pleerock @NoNameProvided do you have any idea about what info and how formated the error message should be? 😕

### DIFF
--- a/src/driver/express/ExpressDriver.ts
+++ b/src/driver/express/ExpressDriver.ts
@@ -278,14 +278,6 @@ export class ExpressDriver extends BaseDriver implements Driver {
                 options.next();
             });
         } 
-        else if (result != null) { // send regular result
-            if (action.isJsonTyped) {
-                options.response.json(result);
-            } else {
-                options.response.send(result);
-            }
-            options.next();
-        }
         else if (result === undefined) { // throw NotFoundError on undefined response
             const notFoundError = new NotFoundError();
             if (action.undefinedResultCode) {
@@ -293,11 +285,19 @@ export class ExpressDriver extends BaseDriver implements Driver {
             }
             throw notFoundError;
         }
-        else { // send null response
+        else if (result === null) { // send null response
             if (action.isJsonTyped) {
                 options.response.json(null);
             } else {
                 options.response.send(null);
+            }
+            options.next();
+        }
+        else { // send regular result
+            if (action.isJsonTyped) {
+                options.response.json(result);
+            } else {
+                options.response.send(result);
             }
             options.next();
         }
--- a/src/driver/express/ExpressDriver.ts
+++ b/src/driver/express/ExpressDriver.ts
@@ -223,18 +223,14 @@ export class ExpressDriver extends BaseDriver implements Driver {
     handleSuccess(result: any, action: ActionMetadata, options: Action): void {
 
         // check if we need to transform result and do it
-        if (this.useClassTransformer && result && result instanceof Object) {
+        if (result && result instanceof Object && this.useClassTransformer) {
             const options = action.responseClassTransformOptions || this.classToPlainTransformOptions;
             result = classToPlain(result, options);
         }
 
         // set http status code
-        if (result === undefined) {
-            if (action.undefinedResultCode) {
-                if (action.undefinedResultCode instanceof Function) {
-                    throw new (action.undefinedResultCode as any)(options);
-                }
-            }
+        if (result === undefined && action.undefinedResultCode && action.undefinedResultCode instanceof Function) {
+            throw new (action.undefinedResultCode as any)(options);
         } 
         else if (result === null) {
             if (action.nullResultCode) {
@@ -290,14 +286,14 @@ export class ExpressDriver extends BaseDriver implements Driver {
             }
             options.next();
         }
-        else if (result === undefined) {
+        else if (result === undefined) { // throw NotFoundError on undefined response
             const notFoundError = new NotFoundError();
             if (action.undefinedResultCode) {
                 notFoundError.httpCode = action.undefinedResultCode as number;
             }
             throw notFoundError;
         }
-        else { // send null/undefined response
+        else { // send null response
             if (action.isJsonTyped) {
                 options.response.json(null);
             } else {
--- a/src/driver/express/ExpressDriver.ts
+++ b/src/driver/express/ExpressDriver.ts
@@ -13,6 +13,7 @@ import {AuthorizationCheckerNotDefinedError} from "../../error/AuthorizationChec
 import {isPromiseLike} from "../../util/isPromiseLike";
 import {getFromContainer} from "../../container";
 import {AuthorizationRequiredError} from "../../error/AuthorizationRequiredError";
+import { NotFoundError } from "../../index";
 const cookie = require("cookie");
 const templateUrl = require("template-url");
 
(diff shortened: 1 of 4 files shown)


src/driver/express/ExpressDriver.ts
```diff
@@ -228,19 +229,24 @@ export class ExpressDriver extends BaseDriver implements Driver {
...
-        } else if (action.nullResultCode && result === null) {
-            if (action.nullResultCode instanceof Function)
-                throw new (action.nullResultCode as any)(options);
-
-            options.response.status(action.nullResultCode);
-
-        } else if (action.successHttpCode) {
+        if (result === undefined) {
+            if (action.undefinedResultCode) {
+                if (action.undefinedResultCode instanceof Function) {
```
NoNameProvided: I know it has been before, but yet this is so bad naming. I would expect it to be number only. We should do some refactoring later.

NoNameProvided:
I have read the code only but it seems ok to me. 

> However I think we should place the info about the route where the undefined has been returned. 

I really think we should add `debugger` package and do logging with proper namespacing. Both Koa and express uses `debugger` as default logger so it wont add dependency overhead, but will make easier to understand to see what happening inside routing-controllers. Some ideas: 

 - adding routes / middlewares
 - returning undefined 
 - errors

Others? I wanted to draft a PR for this quite a while.

19majkel94:
I was thinking about `message` Error property - currently we only have stack and name `NotFoundError`. We should send the info like `Cannot ${action.method} on ${action.url}.` or sth like this. Pure express final handler do sth like this 😉

NoNameProvided:
> I was thinking about message Error property - currently we only have stack and name NotFoundError. 

I dont know how the default errors look like. The first thing was to swap them for custom errors and add an error middleware. :D But yes a `message` property is definitely needed.

pleerock:
hmm why `node: stable` failed?

19majkel94:
@pleerock See https://github.com/pleerock/routing-controllers/issues/243

src/driver/express/ExpressDriver.ts
```diff
@@ -222,25 +223,26 @@ export class ExpressDriver extends BaseDriver implements Driver {
...
             const options = action.responseClassTransformOptions || this.classToPlainTransformOptions;
             result = classToPlain(result, options);
         }
 
         // set http status code
-        if (action.undefinedResultCode && result === undefined) {
-            if (action.undefinedResultCode instanceof Function)
-                throw new (action.undefinedResultCode as any)(options);
-
-            options.response.status(action.undefinedResultCode);
```
pleerock: Looks like you removed this code. Why?

src/driver/express/ExpressDriver.ts
```diff
@@ -275,29 +277,28 @@ export class ExpressDriver extends BaseDriver implements Driver {
...
-                }
-
-                if (action.isJsonTyped) {
-                    options.response.json();
-                } else {
-                    options.response.send();
-                }
-                options.next();
+        } 
+        else if (result != null) { // send regular result
```
pleerock: `!==`

src/driver/express/ExpressDriver.ts
pleerock: Okay looks like it moved to the bottom?

pleerock:
Im afraid of these changes TBH. You changed the most fragile part of code. Are you sure about your changes? Have you tested everything?

src/driver/express/ExpressDriver.ts
19majkel94: `result != null` is an equivalent for `result !== null || result !== undefined`
I should change the order of the checks, so the last `else` would be for `send regular result`

src/driver/express/ExpressDriver.ts
19majkel94: In this place I only left checking `undefinedResultCode`. However now it's not about `set http status code` so I will move it to the handler on the bottom.

19majkel94:
@pleerock It's just a refactor of this part. The only change in behavior is throwing `NotFoundError` when response in undefined to let the error handler format the error info to JSON or text (or custom error handler).

src/driver/express/ExpressDriver.ts
pleerock: yeah it also equivalent of result != 0 and others. Better if you do `result !== null || result !== undefined` if its want you want to do

19majkel94: > yeah it also equivalent of result != 0 and others

nope 😃 

image

pleerock: see? I thought it is! I wanted users to prevent from confusion but was already confused by myself. Please change to explicit checks.

19majkel94: I've already done it (9b701347)+ change order of if-else tree for more understandable code 😉

19majkel94:
> But yes a message property is definitely needed.

@NoNameProvided 
Do you have any proposal of the error message or we should implement the:
`Cannot ${action.method.toUpperCase()} ${action.url}`
like the `finalhandler` in express:
` msg = 'Cannot ' + req.method + ' ' + encodeUrl(parseUrl.original(req).pathname)`
? 😉

NoNameProvided:
Do you ask about the 404 error? We should create error message to every error type. 

I would go with some more general one like `The requested resource doesn't exist.`. I dont see any value adding the url and method to the response, as it is known to the user because he/she made it.

19majkel94:
>  We should create error message to every error type.

Do you mean the default error message for every bulit-in http error subclass? It could be nice 😉 However I always prefer custom messages like:
```ts
const playlistItem = await this.playlistItemRepository.findByIdAndOwner(itemId, user);
if (!playlistItem) {
    throw new NotFoundError(`Playlist item with ID '${itemId}' doesn't exist!`);
}
```
But the default message `The requested resource doesn't exist.` for 404 error is a good idea. So I think we should leave this PR as it is with no params in constructor - the error message would be the default one added later.

So, @NoNameProvided please review this PR, and we will wait for pleerock's acceptation as he is "afraid of these changes TBH" 😉

NoNameProvided:
> Do you mean the default error message for every bulit-in http error subclass?

Yes

>  So I think we should leave this PR as it is with no params in constructor - the error message would be the default one added later.

Yes, its a good idea. It's own my todolist to make error handling swappable / extendable in routing-controllers. I just dont have time to do it.

>  @NoNameProvided please review this PR, 

With just reading the code it looks good to me.

19majkel94:
> With just reading the code it looks good to me.

So please click that damn button! 😆 

image

19majkel94:
@NoNameProvided looks like @pleerock is gone - if you are not afraid of this potentially dangerous changes, please merge this PR 😉

fabiob:
It looks like this PR broke the ability to write directly to the response:

https://github.com/pleerock/routing-controllers#using-request-and-response-objects

19majkel94:
@fabiob 
There are some problems with manually handling the response. You have to wait for this:
https://github.com/pleerock/routing-controllers/issues/117#issuecomment-327232988

There's now a race between routing-controllers action result handling and user's action code. We will fix this soon.

PR #247, opened 2017-08-12
Fix/authorization

NoNameProvided:
Add the ability to throw custom errors in the authorization checker for Koa. 

This is the follow up of #233 to bring parity between the two driver.

It also 

 - adds test to prevent regression of this feature
 - remove forgotten console.logs from the tests

### DIFF
--- a/src/driver/koa/KoaDriver.ts
+++ b/src/driver/koa/KoaDriver.ts
@@ -291,12 +291,10 @@ export class KoaDriver extends BaseDriver implements Driver {
     handleError(error: any, action: ActionMetadata | undefined, options: Action): any {
         if (this.isDefaultErrorHandlingEnabled) {
             const response: any = options.response;
-            console.log("ERROR: ", error);
 
             // set http status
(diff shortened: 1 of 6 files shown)


NoNameProvided:
Wups, had a little ident issue in one of the tests. Fixed now.

NoNameProvided:
So is it good to merge?

NoNameProvided:
By the way, I am not sure about [this line](https://github.com/pleerock/routing-controllers/pull/247/files#diff-112df671cfc9c4c162793234f846b6d0R108). We catch the error and do not rethrow it. I dont know how Koa works to be honest, so I have no idea what it will try to do after we handled the error on our own.

19majkel94:
We capture the error to redirect it to the lib error handler which will do the formating and sending error response. There the `next(error)` is called so everything is ok, right?

NoNameProvided:
Ok, then can we merge? If so please do it, merging own PR-s is not a good thing 😄

PR #252, opened 2017-08-14
Fix/middlewares loading order

### LINKED ISSUE #202: Order of loading of global middlewares
I have two global middlewares (both useBefore), how is the order of execution decided if we load the middlewares from directory.

Secondly, order of middlewares if passed as an array doesn't seem to have any effect. 

```javascript
const app = createKoaServer({
    defaultErrorHandler: false,
    routePrefix: "/v1",
    authorizationChecker: async (action: Action, claims: string[]) => {
        const verifyJwt = jwt(jwtConfig).unless({ path: ['/v1/auth/token', /^\/v1\/accounts/] });
        await verifyJwt(action.context, () => { });
        const user = action.context.state.user;
        if (user.role.includes('admin')) {
            return true;
        }
        for (const claim of claims) {
            if (user.scope.includes(claim)) {
                return true;
            }
        }

        throw new ForbiddenError("you do not have required permissions");
    },
    controllers: [__dirname + "/api/controllers/*.js"],
    middlewares: [CustomErrorHandler, JwtMiddleware]
}).listen(3000);
```

19majkel94:
closes #202.

### DIFF
--- a/src/RoutingControllers.ts
+++ b/src/RoutingControllers.ts
@@ -95,8 +95,7 @@ export class RoutingControllers {
         this.metadataBuilder
             .buildMiddlewareMetadata(classes)
             .filter(middleware => middleware.global && middleware.type === type)
-            .sort((middleware1, middleware2) => middleware1.priority - middleware2.priority)
-            .reverse()
+            .sort((middleware1, middleware2) => middleware2.priority - middleware1.priority)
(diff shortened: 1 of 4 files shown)


19majkel94:
Default priority is 0, so when user mix the ones decorated with priority options and the ones with no priority, it will be stable sorted descending by priority.

NoNameProvided:
Looks good to me.

Issue #218, opened 2017-07-18
Samples should include a test-suite


OysteinAmundsen:
The samples should include a test-runner and unit tests for each controller, demonstrating how one would implement unit testing. I know there are a lot of test frameworks out there, but choose one and a lot of details for all others would fall into place.

This would be a great starting-point for people (like me) who struggles with how to setup i.e. a karma runner for a nodejs typescript project like this.

19majkel94:
@OysteinAmundsen is this your first node.js/express project?

I haven't thought about this as it's even simplier that testing pure express app - you don't have to mock req and res, you just test controller methods like a normal method from services, etc. Just passing inputs params and asserting outputs. With DI you can easily pass mocked service in controller constructor so you don't have to mock and stub the whole world.

I will accept PRs with good samples 😉

gjdass:
Hi.

It seems a great idea. Perhaps I can help on that with pleasure. I'm currently building some sort of white-api skeleton and I'm using routing-controllers as well as typedi and typeorm.
I'm currently building UTs cause I'm lacking of those, but I ran into a strange issue with chai :

The app.ts :
```typescript
private controllers: any[] = [HomeController, UsersController, AuthController];
this.express = createExpressServer({
    routePrefix: "/api/v1",
    controllers: this.controllers
 });
```

HomeController.ts :
```typescript
@JsonController("/")
export class HomeController {

    @Get("/")
    public helloWorld(): any {
        [...]
    }

}
```

UsersController.ts :
```typescript
@JsonController("/users")
export class UsersController {

    @Get("/")
    public async getAll() { [...] }

}
```

So I have 2 routes here, `/api/v1` (default route) and `/api/v1/users` (set thanks to `JsonController` decorator).

Well, chai-http works for requests on the default route, but not on the others.

This **works** (I'm getting into the `.end`) :
```typescript
@test("/api/v1 should return Hello world")
public ShouldReturnHelloWorld() {
    chai.request(expressServer)
    .get("/api/v1/")
    .end((err: any, res: ChaiHttp.Response) => {
        expect(res.status).to.be.equal(200);
        expect(res.body).to.be.equal("Hello world !");
    });
}
```

This **doesn't** (not passing into the `.end`, but the test is green ...) :
```typescript
@test("/api/v1 should return Hello worl")
public ShouldReturnHelloWorld() {
    chai.request(expressServer)
    .get("/api/v1/users") // see here, a JsonController() route
[...]

PR #259, opened 2017-08-20
Simple UTs

gjdass:
2 very simple UTs to start. More may follow.
Express server with routing-controllers.
Testing with mocha-typescript and chai-http if a call to a route returns the right output.

Related (partially) to #218. Some more examples need to be added.

Question tho : is it OK to put package.json and tsconfig.json into that new subfolder ? I hesitated, cause if not, it won't work out of the box cause of packages missing


gjdass:
OK I will answer to myself since Travis is giving me some feedback.
I have moved the local conf of the folder to the global conf of the project.
Hope it's OK.

19majkel94:
I don't see any values of this. You just showed how to use chai-http which has a simple API.
The ask in #218 was about unit testing controllers, not integration/end-to-end endpoints testing.

Please create more advanced controllers with injected services, body params, query params, etc. and show how to test controllers.

gjdass:
Hi. 

Indeed that's quite simple for now. I planned to push others UTs so I guess we gonna wait for further commits before a possible merge.

I'll work on that. 

_Sent from my OnePlus ONE E1003 using [FastHub](https://play.google.com/store/apps/details?id=com.fastaccess.github)_

gjdass:
I've pushed one more UT with Service injection.
I'm learning while coding these (I also do this on a proper project).

Something I don't understand tho : I had to `Container.set()` my service in order to make the thing work, which I don't have into my personal project.
Also I can't figure out how to change my container during tests. If I do it in the static `beforeAll` it's fine, but in tests themselves, nop. Something dumb I don't understand for sure.

pleerock:
I think those changes must be part of separate repository, something like `routing-controllers-test-example` repository which contains test, just like https://github.com/pleerock/routing-controllers-express-demo . Or we even can add changes to that repository instead of creating a new one

NoNameProvided:
@pleerock If I remember correctly you aimed to keep the demo repo simple. Btw I would move the whole sample folder to different repository and add a link here in the readme.

pleerock:
> @pleerock If I remember correctly you aimed to keep the demo repo simple

yeah, thats right. But maybe we can add basic testing functionality as well. But totally not part of this repository.

NoNameProvided:
Lets close this then

Issue #255, opened 2017-08-16
Proposal: New loading controllers/interceptors/middlewares API


19majkel94:
## Introduction
Right now, `RoutingControllersOptions` has following signatures for controllers/interceptors/middlewares -
`Function[] | string[]` - which allows to register in the framework `list of classes or directories from where to import`.

The array syntax is great because it let's you define the order of middlewares explicitly:
```ts
createExpressServer({
    middlewares: [
        MorganMiddleware,
        JwtMiddleware, 
        AuthMiddleware, 
        CompressionMiddleware, 
        LoggingMiddleware, 
        ErrorMiddleware,
    ],
});
```
So there's no need to jump through middlewares files and looking for `priority` option.

However, in big apps it might be not comfortable to list all of 58 controllers explicitly in array, so we have support for glob string and loading from directories:
```ts
createExpressServer({
    controllers: [__dirname + "/controllers/**/*.js"],
})
```
The cons are that you can't disable one middleware for dev/debug purpose and you have to use `priority` option in middleware decorator to declare the order of calling them.

In TypeScript and ES6 we can export and import from modules, so it's a common case to have index files. The directory structure looks like this:
```
middlewares
- auth-middleware.ts
- index.ts
- jwt-middleware.ts
- logging-middleware.ts
// etc.
```
And the index.ts file looks like this:
```ts
export * from "./jwt-middleware";
export * from "./auth-middleware";
export * from "./logging-middleware";
```
And in app.ts:
```ts
import * as middlewares from "./middlewares";
```
Because all object properties are traversed "in the order in which they were added to the object", we can use index.ts to manipulate the explicit order of middlewares. It doesn't have to be placed as index file, it might be placed in app configuration folder with modified paths.

## Main proposal
Right now we can't pass object to routing-controllers option. So we need dirty hacks like:
```ts
createExpressServer(
    middlewares: Object.keys(middlewares).map(key => (<any>middlewares)[key]),
});
```
I propose to add support for objects containing middlewares/controllers/interceptors. We could then just do:
```ts
import * as middlewares from "./middlewares";
createExpressServer(
    middlewares: middlewares,
});
```
or with ES6 shorthand syntax:
```ts
import * as middlewares from "./middlewares";
import * as controllers from "./controllers";
import * as interceptors from "./interceptors";

createExpressServer(
    middlewares,
    controllers,
    interceptors,
});
```
[...]

PR #263, opened 2017-08-24
Support mixing classes & classDirs on RoutingControllerOptions

iDaN5x:
**This change allows us to mix between imported classes & classDirs on:**
- `RoutingContollerOptions.controllers`
- `RoutingContollerOptions. middlewares `
- `RoutingContollerOptions. interceptors ` 

This change also adds `package-lock.json` introduced by npm 5.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,0 +1,5700 @@
+{
+  "name": "routing-controllers",
+  "version": "0.7.1",
+  "lockfileVersion": 1,
+  "requires": true,
+  "dependencies": {
+    "@gulp-sourcemaps/identity-map": {
+      "version": "1.0.1",
+      "resolved": "https://registry.npmjs.org/@gulp-sourcemaps/identity-map/-/identity-map-1.0.1.tgz",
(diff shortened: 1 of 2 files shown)


iDaN5x:
@pleerock node:latest build is failing on all builds.
This PR should not affect lib flow by any means.

iDaN5x:
The proposal seems solid, however, it'll be a breaking change for existing codebases.
While the move to index files is easy enough (hack, it can be automated), #255 is not shipped yet.
I suggest merging my PR as bug-fix, as it is valuable to allow mixing.

19majkel94:
But it's not a bug, it works as designed. And as I said, merging this will make mixing dirs and classes official and I think we (as a developers team) can't support this kind of bad practices.

You can easily do now a workaround by yourself with `as any` type casting.

19majkel94:
> #255 is not shipped yet.

But you can do `Object.values(middlewares)` now, this proposal just make it a official way as a way to depreciate importing from dirs.

iDaN5x:
👍 thanks

pleerock:
dude stop using "depreciate importing from dirs" expression 😆

PR #269, opened 2017-08-30
Koa multipart body

### LINKED ISSUE #156: How to get `multipart/form-data` text field?
I use `@BodyParam` but the value is `undefined`.

fabiob:
This PR fixes #156.

One of the new tests do fail when running under Gulp, but works fine when ran directly from mocha. I have no idea why.

### DIFF
--- a/src/driver/koa/KoaDriver.ts
+++ b/src/driver/koa/KoaDriver.ts
@@ -111,6 +111,8 @@ export class KoaDriver extends BaseDriver implements Driver {
                 .forEach(param => {
                     defaultMiddlewares.push(multer(param.extraOptions).array(param.name));
                 });
+
+            defaultMiddlewares.push(this.fixMulterRequestAssignment);
         }
 
         // user used middlewares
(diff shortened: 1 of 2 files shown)


src/driver/koa/KoaDriver.ts
```diff
@@ -111,6 +111,9 @@ export class KoaDriver extends BaseDriver implements Driver {
                 .forEach(param => {
                     defaultMiddlewares.push(multer(param.extraOptions).array(param.name));
                 });
+
+            // HACK: should be removed when koa-multer merges #15: https://github.com/koa-modules/multer/pull/15
```
NoNameProvided: No need to add the almost same comment twice, remove the comment here.

fabiob: Done

19majkel94:
Ping @NoNameProvided - merge it if you approve it now 😉

NoNameProvided:
Damn, thats why it bad to have failing tests, I assumed only the Cookie test fails on the latest stable but actually this PR brokes the added test in it. @fabiob did this works for you? It seems Koa fails the tests you wrote.

19majkel94:
I can confirm that the test fails on newest node.js. Fortunately, only the newly added test fails so it won't break any existing feature 😆 

Because it fails also in express and not on older node versions, it looks like the second weird test bug on newest node.

Bu the reason why both drivers return 400 Bad Request when using both decorators is that the default decorator option is `required: true` so when it receive undefined, it throw bad request error. And it receive undefined because you CAN't use both `@Body()` and file decorator in one action due to how `multipart/form-data` work - you can send the json only in the field, not as a whole body, because file is there too. So you have to use `@UploadedFile("fileFormName")` and `@BodyParam("jsonFormName")`.

I will push commit with removal of that test and place warning in docs about this.

19majkel94:
@NoNameProvided 
Looks like the multer that is parsing body, is placing files in `req.body` object that doesn't prototypal inherit from Object, so the validator throw empty errors.

I've fixed it in 5c7face, so now you can sleep safe & sound 😉

NoNameProvided:
Thanks.

Issue #123, opened 2017-04-26
Array parameters are not deserialized/validated properly


marshall007:
Given the following decorated parameters on a controller method:

```ts
@QueryParam('fields') fields: string[],
@QueryParam('numbers') numbers: number[]
```

... the behavior of arrays in query parameters is inconsistent. Namely single values are not coerced to an array and the inner values are not coerced or validated against the appropriate type. Some examples:

```js
// GET ?fields=name
// -> ParameterParseJsonError: Value (\"name\") cannot be parsed to JSON

// GET ?fields=foo&fields=bar
console.log(fields) // -> [ "foo", "bar" ]

// GET ?numbers=1&numbers=2
console.log(numbers) // -> [ "1", "2" ] (works, but not deserialized)
```

pleerock:
Right, its not possible to automatically find type of "string[]". You can do it with class-transformer, but your type should be object and inside it you should have your string[] with type set. 

Alternative could be to have `type` option in `QueryParam` options: `@QueryParam("fields", { type: String })`, but this is only needed for arrays, Im not sure if this functionality worth it.

19majkel94:
> You can do it with class-transformer, but your type should be object and inside it you should have your string[] with type set.

So it looks like all the problems with class-validator and class-transformer comes to one solution - #122 😉 I will try to look at the source code to see how much complicated is the implementation of this feature and whether I could help and contribute.

pleerock:
I'll close it. If someone interested in fixing this issue feel free to create a PR. But I personally think it does not worth it. Better to create object and let class-transfromer to perform proper transformation directly into the object, instead of having each array in a separate query parameter.

PR #270, opened 2017-08-31
Fix missing @QueryParams decorator options

### LINKED ISSUE #160: QueryParams does not include ParamOptions as like as the documentation says?
I can not find ParamOptions while using the newly added feature QueryParams?

19majkel94:
This PR add param options for `@QueryParams()` decorator to let enable/disable validation/transforming and other stuffs.

Also it introduce normalizing of query params object properties like it's already done for simple `@QueryParam` decorator.

Fixes #160.


src/ActionParameterHandler.ts
```diff
@@ -103,32 +103,30 @@ export class ActionParameterHandler {
...
-    protected normalizeParamValue(value: any, param: ParamMetadata): Promise<any>|any {
+    protected async normalizeParamValue(value: any, param: ParamMetadata): Promise<any> {
         if (value === null || value === undefined)
             return value;
 
+        // map @QueryParams object properties from string to basic types (normalize)
+        if (param.type === "queries" && typeof value === "object") {
+            Object.keys(value).map(key => {
+                const type = Reflect.getMetadata("design:type", param.targetType.prototype, key);
+                const typeString = typeof type();
```
19majkel94: TBH, I don't like this part, I did this as a quick shortcut.
It may introduce some weird behavior when the reflected type is not constructor-like, however I'm not sure if it's possible in TypeScript? I think I should check if the `type` is undefined for case like no property type provided in class definition.

src/ActionParameterHandler.ts
```diff
@@ -138,8 +138,41 @@ export class ActionParameterHandler {
...
+                if (value === "true" || value === "1") {
+                    return true;
+
+                } else if (value === "false" || value === "0") {
+                    return false;
+                }
+
+                return Boolean(value);
+
+            default:
```
19majkel94: I think we should also cover the `date` case and also the array feature which was requested in #123 - it is possible with class-transform `@Type` decorator to work with arrays.

src/ActionParameterHandler.ts
```diff
@@ -138,8 +138,41 @@ export class ActionParameterHandler {
...
+    }
+
+    /**
+     * Normalizes string value to number or boolean.
+     */
+    protected normalizeValue(value: any, type: string) {
+        switch (type) {
+            case "number":
+                if (value === "")
+                    return undefined;
```
NoNameProvided: I think if a parameter is passed in the query with no value it should be casted to `null` instead of `undefined`.

I will add a normal comment to detail this.

NoNameProvided:
I think if a parameter is passed in the query with no value it should be casted to `null` instead of `undefined`.

`?paramA=3&paramB` should not be equal to `?paramA=3`. For this reason I would return `null` instead of `undefined`. This should also apply to all the other types (boolean, string).

Also this current approach doesn't allow partial description of the queryparams because of the incorrect behaviour of `IsOptional` decorator. (See pleerock/class-validator/issues/101 for details.) Luckily there is a fix for that but it has not been released yet. 

So now if a query class is defined every key must exists, otherwise the validation will fail right?

Example:
```ts
class GetUsersQuery {
 
     @IsOptional()
     @IsPositive()
     limit: number;
 
     @IsOptional()
     @IsAlpha()
     city: string;
 
     @IsOptional()
     @IsEnum(Roles)
     role: string;
 
     @IsOptional()
     @IsBoolean()
     isActive: boolean;
 
 }
 
// this will work for query `?limit=5&city=Amsterdam&role=moderator&isActive=true`
// but wont work for `?limit=5&role=moderator`
```

So to make this work properly we have to pass the following test as well:

```
describe("@QueryParam should give a proper values from request query parameters", () => {
        assertRequest([3001, 3002], "get", "photos?sortBy=name&limit=10&showAll", response => {
            expect(queryParamSortBy).to.be.equal("name");
            expect(queryParamCount).to.be.equal(undefined);
            expect(queryParamLimit).to.be.equal(10);
            expect(queryParamShowAll).to.be.equal(null);
            expect(response).to.be.status(200);
            expect(response).to.have.header("content-type", "text/html; charset=utf-8");
        });
    });
```

NoNameProvided:
Another idea could be to support default params?

Example: 

```ts
class GetUsersQuery {
 
     @IsOptional()
     @IsPositive()
     limit: number = 30
 
     @IsOptional()
     @IsAlpha()
     sortBy: string;
 
     @IsOptional()
     @IsEnum(Roles)
     role: string = Roles.User;
 
     @IsOptional()
     @IsBoolean()
     showAll: boolean = true;
 
 }

describe("@QueryParam should use default values from query descriptor when it's not exist on the requets", () => {
        assertRequest([3001, 3002], "get", "photos?sortBy=name&showAll", response => {
            expect(queryParamSortBy).to.be.equal("name");
            expect(queryParamLimit).to.be.equal(30);
            expect(queryParamShowAll).to.be.equal(null);
            expect(response).to.be.status(200);
            expect(response).to.have.header("content-type", "text/html; charset=utf-8");
        });
    });
```

src/ActionParameterHandler.ts
19majkel94: I've just extracted pleerock's code to separate method, I haven't changed nothing in normalizing logic.
https://github.com/pleerock/routing-controllers/blob/master/src/ActionParameterHandler.ts#L112

I agree that we should convert `""` to true for boolean, but for number we should use undefined if param value not exist or NaN if it's not a number string. However I would like to don't mix different things in one PR, as we first need to discuss the edge cases we should merge this `@QueryParams` PR and start a new one with better normalizing.

19majkel94:
> Also this current approach doesn't allow partial description of the queryparams because of the incorrect behaviour of IsOptional decorator.

It allows just like it allowed before `@IsOptional` was introduced - just use `skipMissingProperties` and `@IsDefined` 😉

19majkel94:
> As you can see this can be easily done in userland, no need for special treatment on framework level.
However I'm not sure if it works well with class-transformer.

Ok, looks like it works - you can define defaults on class level 😉 
```ts
import { plainToClass } from "class-transformer"

class Framework {
    name: string = "type-express"
    version: number = 0.72
    isProductionReady: boolean
}

const test = plainToClass(Framework, {
    isProductionReady: false,
})

console.log(test) // ​​​​​Framework { name: 'type-express', version: 0.72, isProductionReady: false }​​​​​
```

I've added a test case for optional query params with defaults and it works great! You can merge this PR now @NoNameProvided, I will do another one with enhanced params normalization feature soon - this one only brings back the `@QueryParams` decorator options 😉

NoNameProvided:
> No, keep consistent with TypeScript, where optional fields are Type | undefined.

I am still not sure about this. Still my problem is that `?paramA=3&paramB` should not be equal to `?paramA=3` and if we default to undefined it will, because there is no way to tell if `paramB` exists without a value or doesn't exists at all, and its an important information.

So I would like to see: 

- `?paramA=3&paramB=5` --> `{ paramA: 3, paramB: 5 }`
- `?paramA=3&paramB` --> `{ paramA: 3, paramB: null }`
- `?paramA=3` --> `{ paramA: 3 }` (aka `paramB` is `undefined`)

19majkel94:
Variable without value is undefined, not null. Try:
```js
let x;
console.log("x", x);
```
> because there is no way to tell if paramB exists without a value or doesn't exists at all, and its an important information

But your controller want to receive `name` param from query string. What is the difference if your receive `undefined` when param not present and `null` when param is invalid `no string value`?
`GET /user?name` and `GET /user`
In both cases you have no value so you return all values from db, not filtered. [Null is bad](https://basarat.gitbooks.io/typescript/docs/tips/null.html) in JS/TS and we shoudn't use it unless it's necessary (TypeORM deletes fields when value is null, not undefined to allow partial entity updates).

By default, node.js querystring module parse `/user?name` to `{ name: "" }`. I won't normalize empty string to null value for empty string, I can only throw NormalizeError. The discussion is here:
https://github.com/pleerock/routing-controllers/issues/234#issuecomment-326922247

pleerock:
How express reacts if we pass `?paramA=3&paramB` ? What is value of `paramB` in req.query ? is it null or undefined? I think we should do same as what they do

19majkel94:
It is empty string.
https://repl.it/KhlJ

pleerock:
Ah okay it means we cant make decision based on it for numbers, boolean, etc. Im not sure what is correct `null` or `undefined`. For strings it probably must stay "", but for others... @NoNameProvided is there any cases why you need `null` instead of `undefined` ?

19majkel94:
Both `?paramA=&paramB` produces empty string that we have to normalize for boolean to `true` as its common to do e.g. `/todos?showUncompleted`.

pleerock:
Im not sure about such feature. Its very seamless feature

19majkel94:
Ok, but as I said - normalizing query object will be part of another PR. This one only bring back the query params options to let control validation and other options. Could you merge it?

NoNameProvided:
> It is empty string.
https://repl.it/KhlJ

Express doesn't use the querystring package. Or does it? I cannot find it in their package.json.

>  Im not sure what is correct null or undefined. For strings it probably must stay "", but for others... 

I dont get it why you guys want empty string.

> @NoNameProvided is there any cases why you need null instead of undefined ?

I want to make an easily usable difference between sent params with no value and not sent params. When I send a param I do it on a purpose, so I want it to have a different value than undefined which represent uninitialized values.

I get @19majkel94 reasoning that variables without value are undefined, but we should approach this from the side of reading query params not from the side of js. 

Lets take an example: I have a `searchTerm` param.

What I am interested in is that if it's has **meaningful** value. With your proposal I would have to do `if(searchTerm && searchTerm.length > 0)` with my proposal I have to do `if(searchTerm)` because only meaningful values are handled as values. 

If I am interested in seeing if the param was sent without value I can do `if(searchTerm == null)` what explains better for me than `if(searchTerm == "")`

>  Both ?paramA=&paramB produces empty string that we have to normalize for boolean to true as its common to do e.g. /todos?showUncompleted.

I am ok with this. It's a nice trick, just needs to be outlined in the docs.

19majkel94:
> Express doesn't use the querystring package. Or does it? I cannot find it in their package.json.

It's a built-in node module. And they use "qs" package which works in the same way:
image

> I want to make an easily usable difference between sent params with no value and not sent params. When I send a param I do it on a purpose, so I want it to have a different value than undefined which represent uninitialized values.

It doesn't work like this in node.js world as you see. So you can't change the world, we just normalize the params like converting "true" to true or "123" to 123.

So as both `?paramA=&paramB` produces empty string, we should map the empty string to `true` for boolean, for number throw normalizing error and leave it as it is for string params.

> What I am interested in is that if it's has meaningful value. With your proposal I would have to do if(searchTerm && searchTerm.length > 0) with my proposal I have to do if(searchTerm) because only meaningful values are handled as values.

Have you forgotten that `""` and `0` evaluates to false?
```ts
const str = ""
if (str) {
    console.log("empty string?")
} else {
    console.log("no, conversion to truthy values happen") // logs this
}
```

NoNameProvided:
> Have you forgotten that "" and 0 evaluates to false?

Damn, forgotten about "" 😱 

Yet I am not convienced on the string one. But it's 2 vs 1 so I guess we will go that way.

19majkel94:
> If I am interested in seeing if the param was sent without value I can do if(searchTerm == null) what explains better for me than if(searchTerm == "")

But what business logic depends if the string param has been set but with no value. Like:
```
GET /users?city
```
```ts
if (city === undefined) { // city param not exist
   return usersService.getAll();
} else if (city === null) { // city param exist but has no value
   return usersService.getUsersWithCityDefined(); // ???
}
```
For boolean no value means true, not exist means undefined, right?
For numbers no value means no number, so undefined. If it's required we have then `ParamRequiredError`.
So as I explained, for string we signalize the "exist but with empty value" by empty string. No need to pollute by setting value to null which breaks TS type system.

19majkel94:
Closing in favor of #289.

PR #274, opened 2017-09-04
fix(ExpressDriver): use response.render to render templates

NoNameProvided:
Re-implement #225 and add test for it.


NoNameProvided:
So, ok to merge?

PR #277, opened 2017-09-05
Handle different content-type responses in JsonController

### LINKED ISSUE #236: Using @Controller with @ContentType throws me an error
Hi,

I have managed to get my endpoints working using @JsonController, works flawlessly. I have now changed over to @Controller, the reason being is that one of the endpoints needs to be something other than application/json.

So I have decorated my method (Its a POST) with 

       @ContentType("application/json"

Using postman I initiate a call but receive "Unexpected 'E'"

and the node session crashes with the following

> Error
>     at BadRequestError.HttpError [as constructor]

Removing the ContentType decorator and changing back to JsonController works.

I can confirm I am calling in using Content-Type: application/json 

Anyone know whats going on ?

19majkel94:
Fixes #236.

### DIFF
--- a/src/metadata/ActionMetadata.ts
+++ b/src/metadata/ActionMetadata.ts
@@ -174,6 +174,7 @@ export class ActionMetadata {
         const redirectHandler = responseHandlers.find(handler => handler.type === "redirect");
         const renderedTemplateHandler = responseHandlers.find(handler => handler.type === "rendered-template");
         const authorizedHandler = responseHandlers.find(handler => handler.type === "authorized");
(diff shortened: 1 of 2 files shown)


19majkel94:
Should we implement support for `@ContentType`  on controller class? This could be the default content-type which could be overwritten by the action decorator. It might be helpfull when the controller has many endpoints returning content of selected type (e.g. "text/html") as right now `@JsonController` behave just like that.

pleerock:
Yeah I think we can do that as well

PR #182, opened 2017-06-13
Stream type support for express driver

Luke265:

PR #285, opened 2017-09-08
Add support for returning Buffer and streams from action

### LINKED ISSUE #180: Pipe stream to response returns 404
How to send a stream?

````
import {Get, JsonController, Res} from "routing-controllers";

@JsonController('/test')
export class TestController {

    @ContentType('application/octet-stream')
    @Get('/stream')
    stream(@Res() res) {
        require('fs').createReadStream('someFile.dat').pipe(res);
    }

}
````

The example above does not work correctly, returns 404, unless we remove https://github.com/pleerock/routing-controllers/blob/master/src/driver/express/ExpressDriver.ts#L301 line.

### LINKED ISSUE #183: Support returning Stream from controller methods
When the user returns a `Stream` from their controller method, we should automatically pipe it to the response. This would fix #180 and make it much more obvious how to properly implement that use-case.

### LINKED ISSUE #228: how to return a binary file response
I am returning a binary file as a response, but it is returned as a json:

```javascript
{
    "0": 37,
    "1": 132,
    "2": 114,
    "3": 119,
    "4": 150,
    "5": 148,
    "6": 170,
    "7": 246,
    "8": 138,
    ...
}
```
I am using `@Controller` decorator and tried setting the content-type header but content-type is set to `application/json` even if I set it explicitly to `application/octet-stream` but still it didn't work:

```javascript
@Get("/:id/product.dat") 
@Header("Content-type", "application/octet-stream") // didn't work
async getProductFile( @Ctx() ctx: any, @Param("id") id: number) {
        ...
        const productFileDat: Buffer = hexDecode(productFile);
        ctx.set("Content-Type", "application/octet-stream");  // didn't work either
        ctx.attachment("Product.dat");
        return productFileDat;
 }
```

19majkel94:
Reimplementing PR #182.
Fixes #228, fixes #180 and fixes #183.

### DIFF
--- a/src/driver/BaseDriver.ts
+++ b/src/driver/BaseDriver.ts
@@ -1,5 +1,5 @@
 import {ValidatorOptions} from "class-validator";
-import {ClassTransformOptions} from "class-transformer";
+import {ClassTransformOptions, classToPlain} from "class-transformer";
 
 import {HttpError} from "../http-error/HttpError";
 import {CurrentUserChecker} from "../CurrentUserChecker";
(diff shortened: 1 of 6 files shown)


fabiob:
@19majkel94, I'm bringing the discussion from https://github.com/pleerock/routing-controllers/issues/228#issuecomment-328265240 to here.

Here's how I test binary downloads on my app:

https://gist.github.com/fabiob/70b6d7d3022dc6fa36b01ce1d60a5de4

19majkel94:
I've tested it manually by creating sample server and sending file buffer as a response. PDF and mp3 works great, so looks like there's no need to change.

acuntex:
Works great, thank you @19majkel94

PR #297, opened 2017-09-23
Patch to treat unhandled rejections as test failures

fabiob:
We have some bugs on `routing-controllers` codebase that are somewhat silent: they just log an unhandled rejection. Since that's deprecated and undesirable (future node versions will terminate the process on unhandled rejections), I've augmented `assertRequest` so it fails tests when it happens.

The `assertRequest` augmentation made another bug surface (expressjs/express#2116), so I fixed it on this PR as well.


pleerock:
Its really hard to make decision to merge something for an issue you haven't faced with. Okay lets do it

PR #318, opened 2017-10-25
Expose error name independently of error message

twittwer:
Hi,
in my opinion the ability to interpret the name (type) of an error in the frontend/client would be useful.
I don't understand why the exposure of the name should depend on the existence of an error message.
That's why I would appreciate if the name would be exposed, if it specific defined.

My current version is still invoking the existence of an error message:
```javascript
if (error.name && (error.name !== "Error" || this.developmentMode || error.message))
    processedError.name = name;
```
If you agree with the independence, I would propose:
```javascript
if ((error.name && error.name !== "Error") || this.developmentMode)
    processedError.name = name;
```

Best Regards


twittwer:
@19majkel94 Thx, would you agree to expose if by default - without any condition?

19majkel94:
I don't get why we print error name in production only if `error.message` exist. So it might happen that something went wrong on on the server and we get error code 500 with no info in response body 😆

twittwer:
Ok, I've removed the condition :slightly_smiling_face:

NoNameProvided:
@19majkel94 can we squash merge for cleaner history? Or just simple merge?

19majkel94:
@NoNameProvided I think yes, it should not be the problem with this PR case.

Issue #226, opened 2017-07-20
Performance Issue with big json result


asaieed:
Hello Guys,

The response time take too much time in order to return the response of a big json result.
while tracing the issue i get the result from database in about 1.2 sec as a json result, then after return it to action method takes too much time to resolve.

I tried using @Controller() instead of @JsonController() in order to not forcing the response to re-parse again, with no effect.

The only solution worked with me is, getting the KOA context "@Ctx()" and send the result directly, and it's working fine now.

Any other solution for this issue?

NoNameProvided:
Can you do some perf metrics to see where the majority of the time is spent? One quick advice, you should try to skip circular check in class-transformer

asaieed:
Hello @NoNameProvided, I did that as i mentioned and found that i have the result in memory after about 1.2 sec and after return the data to action method it takes this much time to respond the request.

PS: how can i skip circular check in class-transformer?

NoNameProvided:
The only way is to disable `class-transfomer` when you create your server via the `classTransformer: false` option. However this disables `class-transfomer` entirely (so no typed paramters) not just the `skipCircularCheck`. It is ok to turn it off for testing this issue, but probably a bad idea to turn it of completely.

There have been some discussion about adding an option to disable circular check recently, but seems it haven't been implemented yet. If disabling `class-transfomer` solves your problem you can do more research by cloning the class-transformer repo and calling plainToClass by hand on your test data.

asaieed:
This is so cool, but despite talking about "class-transfomer" what is the difference between "@Controller" and "@JsonController", I think the first one should return the result as it is without re-parsing it to JSON because in my case its actually JSON, this is because after sending the result directly to KOA @Ctx the problem has been solved.

NoNameProvided:
Both `@Controller` and `@JsonController` is used to signal that you want to register each decorated controller method (with `@Get`, `@Post`, etc...) in that given class as a route handler. 

When you use `@JsonController` then all controller actions will return serialized json data, and the response's  `content-type` always will be `application/json`. A basic `@Controller` will do nothing with the returned value.

19majkel94:
> A basic @Controller will do nothing with the returned value.

Are you sure?
```ts
if (options.classTransformer !== undefined) {
    driver.useClassTransformer = options.classTransformer;
} else {
    driver.useClassTransformer = true;
}
```
```ts
// check if we need to transform result and do it
if (this.useClassTransformer && result && result instanceof Object) {
[...]

Issue #179, opened 2017-06-09
use classTransformer only for input


jselesan:
Hi there. A time ago, I had to disable classTransformer (in useExpressServer) because it didn't work fine with mongoose models, but now, I would like to take advantage of classTransformer to transform the query params to an object, so it would be nice to have the ability to enable classTransformer only for the input and keep it disabled for output. Is it posible?

pleerock:
no I don't think its possible right now. What problem do you have with mongooose models? Maybe `@Exclude` can help you in your mongoose models?

ruiming:
It would be nice to have a option to disable transform on response.
Return a sequelize instance or mongoose instance both throw error.

I now doing this to disable transform on response:

```typescript
import { KoaDriver } from 'routing-controllers/driver/koa/KoaDriver'

const originHandleSuccess = KoaDriver.prototype.handleSuccess
KoaDriver.prototype.handleSuccess = function () {
  this.useClassTransformer = false
  return originHandleSuccess.apply(this, arguments)
}
```

pleerock:
thats why you should use typeorm instead of sequelize or mongoose 😝 . But seriously it should be easily fixed if you do something like this in your models:

```ts
@Exclude()
_id: string;

@Expose()
get id() {
    return this._id.toString(); // or some other method that normalizes that id
}
```

19majkel94:
Closing as a duplicate in favor of #266.
Please subscribe there to be notified when the required changes land.

AnimaMundi:
@ruiming Nice workaround, thanks!

@pleerock I really appreciate the work you've done with routing controllers so far, it's great! But I believe a library like this should be agnostic of any data mappers being used rather than being developed to suit only TypeORM.

pleerock:
It does not "suit typeorm". class-transformer is generic serializer and does work with object. In mongodb you have complex MongoID object and it simply tries to serialize it as well. The issue with complexity of mongoid object, not with class-transformer.

AnimaMundi:
> thats why you should use typeorm instead of sequelize or mongoose

I was referring to your comment, which implies you shouldn't be using anything other TypeOrm with routing-controllers.

pleerock:
I didnt say that routing-controllers cannot be used with anything else except typeorm. It was simple advice and a bit of joke as you can see from "😝" smile I used.

AnimaMundi:
Okay that's my bad 😆 sorry for misunderstanding!

pleerock:
[...]

Issue #149, opened 2017-05-19
Problem with mongoose model serialization


jselesan:
Hi!. I'm using routing-controllers with mongoose, but when my controller returns the result of a query, the _id field of the models are serialized as

```
{"_id":{"_bsontype":"ObjectID","id":{"0":89,"1":17,"2":175,"3":60,"4":137,"5":110,"6":193,"7":50,"8":29,"9":2,"10":180,"11":121}}
```
My method is:
```
@Get("/")
  all() {
      return this.usersService.getAll().then( users => users.map(u => u.toJSON()) );
  }
```

I think that the problem is the class transformer, but I don't know how to fix it

jselesan:
I'm sorry. Just disabled the classTransformer on the options and all works fine for me.

PR #301, opened 2017-10-02
Controller inheritance

ivanpadavan:
This PR aims to fix #147.

First of all, thank you for amazing typestack! I've just started my exploration and got a feeling that I found a robust solution for full-stack typescript programming. This solves may be more concerns that I have at the moment, so huge thank you again. Making backend was never so nice and natural.

I found a problem that bothers me - inability to use inheritance without limitations & concerns.I've found it's unnatural and counterintuitive.

Solving this provides an opportunity to create complicated controllers (that simulates OData endpoint, for example) which could be reused easily. Or the definition of simple CRUD controller becomes a one-liner. Which is great, as for me. So solving this also opens doors for contribution on top of the stack.

Best Regards, sorry for my english.

PR #329, opened 2017-11-22
Add a useResponseClassTransformer global option

epiphone:
Hi, first of all thanks for a great library!

By default response values are coerced to plain objects with [class-transformer](https://github.com/pleerock/class-transformer). When returning large objects (#226) or values with complex serialization logic (e.g. Mongoose documents #149) you might opt for the default `toJSON` handler instead.

This PR adds a global `useResponseClassTransformer` option parameter for toggling response transformation. It defaults to `true`, and is overridden by `classTransformer: false`.

Also related to #179.


src/driver/BaseDriver.ts
```diff
@@ -129,14 +135,14 @@ export abstract class BaseDriver {
 
     protected transformResult(result: any, action: ActionMetadata, options: Action): any {
         // check if we need to transform result
-        const shouldTransform = (this.useClassTransformer && result != null) // transform only if enabled and value exist
+        const shouldTransform = (this.useClassTransformer && this.useResponseClassTransformer && result != null) // transform only if enabled and value exist
```
19majkel94: not updated comment

19majkel94:
Isn't that the feature that @NoNameProvided promised to implement a few months ago? 😉

NoNameProvided:
Yes, it was. Shame I haven't had the time. I haven't reviewed this either because I think we had agreed on a slightly different approach, but I am not sure, and I don't have time to read it back now. I have a lot of thing on my table now, so I struggle to find any time to spend on my hobbies (including this lib) right now.

epiphone:
Hi, I updated the PR to better match with @NoNameProvided's solution [outlined here](https://github.com/typestack/routing-controllers/issues/226#issuecomment-317025358), i.e. added `transformRequest` and `transformResponse` options to controller and route decorators. Gotta agree this approach seems more future-proof :smile: 

What do you think?

src/decorator-options/ControllerOptions.ts
```diff
@@ -0,0 +1,8 @@
+import {TransformationOptions} from "../TransformationOptions";
+
+/**
+ * Extra options that apply to each controller action.
+ */
+export interface ControllerOptions extends TransformationOptions {
```
19majkel94: I think that you should implement `TransformationOptions`, not extending it - it would be ready for future extension/changes because you can't extend multiple interfaces.

test/functional/action-options.spec.ts
```diff
@@ -0,0 +1,70 @@
...
+    before(done => expressApp = createExpressServer().listen(3001, done));
+    after(done => expressApp.close(done));
+    before(done => koaApp = createKoaServer().listen(3002, done));
+    after(done => koaApp.close(done));
+
+    describe("use controller options when action transform options are not set", () => {
+        assertRequest([3001, 3002], "post", "default", { firstName: "Umed", lastName: "Khudoiberdiev" }, response => {
+            expect(initializedUser).to.be.instanceOf(User);
+            expect(response).to.have.status(200);
+            expect(response.body.lastName).to.be.defined;
```
19majkel94: You've set `transformResponse: false` but you expect `lastName` to be defined even if `lastName is excluded when class-transformer is disabled`... I don't get it 😕

test/functional/controller-options.spec.ts
```diff
@@ -0,0 +1,88 @@
...
+        getMetadataArgsStorage().reset();
+
+        @JsonController("/default")
+        class DefaultController {
+            @Post("/")
+            postUsers(@Body() user: User) { return handler(user); }
+        }
+
+        @JsonController("/transform", {transformRequest: true, transformResponse: true})
+        class NoResponseTransformController {
```
19majkel94: `NoResponseTransformController`? **No?** 😆

test/functional/controller-options.spec.ts
```diff
@@ -0,0 +1,88 @@
...
+    before(done => expressApp = createExpressServer().listen(3001, done));
+    after(done => expressApp.close(done));
+    before(done => koaApp = createKoaServer().listen(3002, done));
+    after(done => koaApp.close(done));
+
+    describe("controller transform is enabled by default", () => {
+        assertRequest([3001, 3002], "post", "default", { firstName: "Umed", lastName: "Khudoiberdiev" }, response => {
+            expect(initializedUser).to.be.instanceOf(User);
+            expect(response).to.have.status(200);
+            expect(response.body.lastName).to.be.defined;
```
19majkel94: Again:
- default: `...lastName).to.be.defined;`
- enable: `...lastName).not.to.be.undefined;`
What's going on? Isn't transform enabled by default?

test/functional/action-options.spec.ts
```diff
@@ -0,0 +1,70 @@
...
+        });
+    });
+
+    describe("override controller options when action transform options are set", () => {
+        assertRequest([3001, 3002], "post", "override", { firstName: "Umed", lastName: "Khudoiberdiev" }, response => {
+            expect(initializedUser).not.to.be.instanceOf(User);
+            expect(response).to.have.status(200);
+            expect(response.body.lastName).to.be.defined;
```
19majkel94: Shouldn't you use `.exist` here? I can't see `defined` in chai documentation 😕

epiphone:
Thanks for the feedback @19majkel94, lots of good points in there :+1: . I refactored the tests, let me know if there's anything else you'd like to see changed.

I did have some trouble with class-transformer's global metadata storage: calling `defaultMetadataStorage.clear()` ([here](https://github.com/typestack/routing-controllers/blob/master/test/functional/class-transformer-options.ts#L32) for example) deletes the class-transformer metadata of all defined classes, so I had to move the `User` class definitions inside a `before`-block and [assign it to a variable](https://github.com/typestack/routing-controllers/pull/329/commits/c4089663728183e965f78ba226abddf3171192c1#diff-d0a8cf7e0ad5016427a3046b660399fcR36) in the outer scope. The end result is still reasonably clean, I think, but let me know if you'd have a better solution.

19majkel94:
Yeah, it clears because you call `clear` 😆 But I don't see the need to do this - pleerock does only:
```ts
 before(() => {
            getMetadataArgsStorage().reset();
```
In class-transformer-options tests. So you should only once declare the model and clear only the routing-controllers metadata.

epiphone:
Because the way it was set up, after a test module finished it would not only clear the metadata of its own test classes, but also of all the classes declared in the following tests. For example after `class-transformer-options.ts` finished, it wiped out `User`'s metadata in `global-options.spec.ts` which would break the test. That's why I ended up delaying the class declarations until the `before` block; I'd be happy to refactor it though if you have an alternative solution in mind.

19majkel94:
So the problem was that pleerock uses the class-transformer only in one test so clearing metadata wasn't affect other tests but now you have tests also in other files so it clears the metadata of all tests, right? 😄

epiphone:
Exactly :+1: . I took a peek at class-transformer tests, seems like they're handling things in a similar way, declaring classes in the upper `describe` and `it` blocks.

19majkel94:
Ok, you have `user.lastName || "default";` in your tests but I can't see a test case that is using it. Please make a one more test case that have `transformRequest: true` and `transformResponse: false` to see that it's working and we are ready to merge 😉

epiphone:
Hey, sorry for the delay, got kinda busy over the holidays! I added the test case and removed `lastName = "default";` where it's unused.

19majkel94:
I'm not sure if @pleerock will like the options object inside decorator so I will wait for his review before merging this PR 😉

vekexasia:
Hello, i'm jumping here to understand if this was holded back due to an alternative way to use both JSON controllers and "Raw" controllers that returns buffers

epiphone:
I'm also curious about what's holding this back. Care to comment @NoNameProvided ?

f4z3k4s:
@pleerock, @NoNameProvided could you please have a look at this? This is still an existing issue :(

kunokdev:
Is there anything happening regarding this?

CSLTech:
Is there still something holding this back @NoNameProvided @pleerock ?

MohammedSiddiqui:
What's the update on this ? :sweat_smile:

hallsamuel90:
Also curious about the status of this?

jotamorais:
@hallsamuel90, @MohammedSiddiqui @epiphone, @CSLTech, @kunokdev, @f4z3k4s, @vekexasia, @MichalLytek.

I'm planning to merge this PR and release this tomorrow night among some other fixes:

- #543 fix(Interceptors): order of global interceptors 
- #518 Input-validation bypass vulnerability 
- #438 class-validator arbitrary bypass vulnerability
- #536 Add support for wildcard "all" routes
- #546 chore: add another example for using the response directly
- #329 Add a useResponseClassTransformer global option
- #226 Performance Issue with big json result
- #149 Problem with mongoose model serialization 
- #179 use classTransformer only for input
- #301 Controller inheritance
- #147 Controller Inheritance support

yurkagon:
Still waiting :)

PR #331, opened 2017-11-23
Param name in BadRequestError

### LINKED ISSUE #330: QueryParam name not known when validation fails
hi,

lets say we have an action with two query params

```
@Get('/')
anAction(
      @QueryParam('filter', {validate: true}) filter: SomeFilterObject,
      @QueryParam('sort', {validate: true}) sort: SomeSortObject
){
 // ...some fancy logic here
}
```

now when the validation fails on either of the params, we can never tell which one failed.
thats all because when throwing `BadRequestError` in
https://github.com/pleerock/routing-controllers/blob/master/src/ActionParameterHandler.ts#L191
it only contains the validation errors itself, and no information about the parameter it failed the validation on.

My idea is to have the error carry information about parameter name at least. this way in a custom error middleware it is possible to catch that error and provide proper information what went wrong with validation.

sh3d2:
Parameter name in BadRequestError

closes #330

### DIFF
--- a/src/ActionParameterHandler.ts
+++ b/src/ActionParameterHandler.ts
@@ -190,6 +190,7 @@ export class ActionParameterHandler<T extends BaseDriver> {
                 .catch((validationErrors: ValidationError[]) => {
                     const error: any = new BadRequestError(`Invalid ${paramMetadata.type}, check 'errors' property for more info.`);
                     error.errors = validationErrors;
+                    error.paramName = paramMetadata.name; 
                     throw error;
                 });
         }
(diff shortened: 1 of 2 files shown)


sh3d2:
a PR for #330

NoNameProvided:
Hi @sh3d2! 

Thanks for your contribution, I think we can add this, but before I can merge plese add some test covering this feature.

sh3d2:
Hi @NoNameProvided,

Thanks for a quick response. ive added the tests you asked for;)

sh3d2:
@19majkel94 no, actually it works ootb, ive simplified the tests (no more custom error handler)

sh3d2:
@19majkel94 @pleerock ping, id very much like to push this one forward.

19majkel94:
@NoNameProvided please check this PR and merge it 😉

PR #359, opened 2018-01-26
Update class-validator dependency

Bajdzis:
I must use @ValidateNested() (fix in class-validator  v0.8.0) with @Body({validate: true})


jotamorais:
Closing the PR as it has been fixed through https://github.com/typestack/routing-controllers/pull/512

Issue #391, opened 2018-04-10
Providing any ValidationOptions in @Body overwrites entirety of defaults set at server level


d0ck:
If I have set ValidationOptions at the server level thus so:

```
this.app = createKoaServer({
      validation: {
        forbidNonWhitelisted: true, // error when fields not defined in validation class are present
        forbidUnknownValues: true,
        skipMissingProperties: true, // this gives us more fine grained control via groups
        whitelist: true, // strip all fields not defined in validation class
      },
});
```
but then provide local validation options in the `@Body()` decorator (or I'm assuming anywhere else local `validate: boolean | ValidationOptions` is accepted):

```
@Post()
public async somePostEndpoint(
  @Body({
      validate: {
        groups: ["myGroup"],
      },
  })
  representation: RepresentationData,
) {
    // endpoint code
}
```
When that endpoint is hit it does not use any of the server level defaults. This is because in the `ActionParameterHelper` the code to figure out which validation options to use looks like this:

`var options = paramMetadata.validate instanceof Object ? paramMetadata.validate : this.driver.validationOptions;`

Now, this could be by design, however I as a user I was assuming I would only be overwriting ValidationProperties which I provided locally, rather than the entirety of my server level defaults. This makes it more complicated to use localized options like groups which could be different for every endpoint. I propose a merge that prioritizes the local options provided (i.e. in the `@Body()` decorator).

`var options = Object.assign({}, this.driver.validationOptions, paramMetadata.validate);`

I can put that in a PR if accepted.

choyg:
Running into this as well. Either need to merge PR or provide explicit docs because this is a glaring security hole.

PR #392, opened 2018-04-11
Fix: Providing any ValidationOptions in @Body overwrites entirety of defaults set at server level #391

d0ck:
Added the change described in the issue I opened #391 as well as a test to cover.


nolazybits:
People we need movement please!
So many PRs opened bringing well and long wanted fixed being forgotten...
This is a must have too!

nolazybits:
@d0ck thanks for the PR.
We have starting creating forks of some of the typestack libraries that are stale here
https://github.com/ts-revival

We have created a PR based on your repo/branch and merged it
https://github.com/ts-revival/routing-controllers/pull/2

Thanks again

PR #418, opened 2018-07-04
Handle stream errors in Express driver

bvanreeven:
When the result of an action is a stream, register an error handler on the stream, so that the appropriate Express error handler will be called. Without such an error handler, an error on the stream will make the Node process crash.

For Koa this is not needed, as it automatically registers an error handler when returning a stream: https://github.com/koajs/koa/blob/74170caf0b332dae01fb81cf03145a17df83b6dd/lib/response.js#L170


bvanreeven:
@19majkel94 The Travis CI build failed, but those failures do not seem related to my changes.

Node stable: `No output has been received in the last 10m0s`
Node 8: succeeded
Node 6: `node_modules/@types/koa/index.d.ts(24,57): error TS2307: Cannot find module 'http2'.`

Issue #428, opened 2018-08-15
kcors has been renamed to @koa/cors


acuntex:
`useKoaServer `still uses the package `kcors` when setting `cors` to `true`.

The package `kcors` has been recently renamed to `@koa/cors`.

daveabes:
Excuse me, but will the configuration of kcors remain the same as what is displayed on the guthub manual?

PR #429, opened 2018-08-15
implemented compatibility to @koa/cors

acuntex:
According to https://github.com/typestack/routing-controllers/issues/428


src/driver/koa/KoaDriver.ts
```diff
@@ -43,7 +43,7 @@ export class KoaDriver extends BaseDriver {
         const bodyParser = require("koa-bodyparser");
         this.koa.use(bodyParser());
         if (this.cors) {
-            const cors = require("kcors");
+            let cors = require("@koa/cors") || require("kcors");
```
19majkel94: What is the behavior of `require` when module doesn't exist? It throws error or return falsy value?

PR #432, opened 2018-08-25
subdomain support koa only

Keyj1n:
It is now possible to declare a controller to a specific subdomain.

```typescript
    import {Controller, Get} from "routing-controllers";

    @Controller({ subdomain "service" })
    export class ServiceController {

        @Get("/")
        get() {
           return "This action is only fired when on subdomain service";
        }
    }

    @Controller({ subdomain "*.service" })
    export class CustomerServiceController {

        @Get("/")
        get() {
           return "This action is only fired when on subdomain *.service";
        }
    }

    @Controller({ baseRoute: "/api", subdomain "*.service" })
    export class CustomerServiceController {

        @Get("/")
        get() {
           return "This action is only fired when on subdomain *.service and route /api";
        }
    }
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -97,7 +97,7 @@ You can use routing-controllers with [express.js][1] or [koa.js][2].
 
     **b. If you want to use routing-controllers with *koa 2*, then install it and all required dependencies:**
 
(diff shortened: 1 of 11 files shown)


Keyj1n:
can somebody please review this?

PR #455, opened 2018-11-25
Fix/Feature: adding support to manually handle responses

aviranco:
The current implementation does not support manually setting the response, via .json(), .end() etc.
1. The library always called response.end() after the action finished event hough it might already been set, and throwing error "cannot set headers after sent".
2. If setting the response in process.nextTick or other async manner, the action method will exit and set the response before the developer get a chance to set it. 
3. Additionally, middleware did not called correctly after the manual response send, with content=undefined, since they were not aware to the manual response, and even called before the response was ready.

I've added support to manually set the response and fix those issues.

To tell the action you want to manually set the response, use:
```
getItems(@Res({ manual: true }) response :ManualResponse) {
    process.nextTick(async () => {
    response.json({message: 'hey bro'});
    });

    return response.manual();
  }
```

I am throwing indicative errors when stating `manual:true` but not returning `response.manual()`, or calling `response.manual()` when not stated `manual:true`

I can also implemented it to simply return response or simply ignore the return value, but I wanted to make it explicit so whoever looks on the return value, he will be aware that it is set manually in other place.


19majkel94:
You should ask in the issue first - this feature already exist and is described in docs:
https://github.com/typestack/routing-controllers#using-request-and-response-objects
> If you want to handle the response by yourself, just make sure you return the response object itself from the action.

https://github.com/typestack/routing-controllers/blob/4a56d176db77bc081dfcd3d8550e8433b5bc476e/src/driver/express/ExpressDriver.ts#L241-L245

And about this problem:
> Additionally, middleware did not called correctly after the manual response send, with content=undefined, since they were not aware to the manual response, and even called before the response was ready.

You're right but the better way would be to create a `@Next()` decorator to call the next function rather than patching the `response.prototype.end` function to detect when to call the next function.

PR #472, opened 2019-02-14
make class-validator a peer dependency

### LINKED ISSUE #384: Routing-Controllers + class-validator bug?
> From 	[11000100111000](https://github.com/11000100111000) in [typestack/class-validator#128](https://github.com/typestack/class-validator/issues/128)

When using the automatic validation from routing-controllers for some reason the email "patch@patch" is valid when we are using these validators and the "patch" validator group
```js
@IsEmail({
        allow_display_name: false,
        allow_utf8_local_part: true,
        require_tld: true
    }, {
        groups: ["post", "patch", "login"]
    })
    @IsOptional({
        groups: ["patch"]
    })
    public email;
```
Doing it the manual way like so works however
```js
let user = new UserValidation();
user.email = "patch@patch";
validate(user, {groups: ["patch"]}).then(console.log);
```

sincraianul:
closes #384

It makes sense to have class-validator as a peer dependency since routing-controllers is not really relying on it to provide certain standalone functionalities, but rather executes code from class-validator on the developer's part.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "routing-controllers",
-  "version": "0.7.7",
+  "version": "0.7.8",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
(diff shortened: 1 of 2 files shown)


matthewlilley:
It would be appreciated if you can accept this PR, or make a commit to achieve this yourselves. This causes problems when using class validator in other areas of an application.

sincraianul:
@matthewlilley FIY I've transitioned to NestJS meanwhile, as this repo doesn't seem to be actively supported.

matthewlilley:
@sincraianul  Thanks for that, NestJS looks impressive, and resembles a popular PHP rad framework, which is something I believe the JS eco-system has been desperately needing for some time.

This wasn't a blocker for me, I just rolled back to the same version of class validator as this package is using, however, it would be nice to not be locked into using this version.

jotamorais:
Fixed and merged through #512

PR #478, opened 2019-03-17
fix: fix options type

unix:
`fileUploadOptions` is `object`.


unix:
Reopen to waiting for a reply.

PR #484, opened 2019-04-11
fix(driver): reject multiple calls

### LINKED ISSUE #220: One url make 2 action call

unix:
Reject action duplicate calls caused by unsatisfactory design.
About issue: #220 .

fixed #220 .

### DIFF
--- a/sample/sample1-simple-controller/UserController.ts
+++ b/sample/sample1-simple-controller/UserController.ts
@@ -20,12 +20,17 @@ export class UserController {
             { id: 2, name: "Second user!" }
         ];
     }
+  
+    @Get("/users/1")
+    getSpecialOne(@Req() request: Request) {
+      return "User 1, Special.";
+    }
 
     @Get("/users/:id")
     getOne(@Req() request: Request) {
         return "User #" + request.params.id;
     }
-
+    
     @Post("/users")
     post(@Req() request: Request) {
         let user = JSON.stringify(request.body); // probably you want to install body-parser for express
--- a/src/driver/express/ExpressDriver.ts
+++ b/src/driver/express/ExpressDriver.ts
@@ -246,7 +246,6 @@ export class ExpressDriver extends BaseDriver {
      * Handles result of successfully executed controller action.
      */
     handleSuccess(result: any, action: ActionMetadata, options: Action): void {
-    
         // if the action returned the response object itself, short-circuits
         if (result && result === options.response) {
             options.next();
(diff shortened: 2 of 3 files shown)


Elassyo:
This could also be fixed by appending an empty handler to the list of middlewares when registering the action in the driver so that express stops handling the request.
Also, is this project still maintained? This PR is 10 months old without any response from anyone but bots...
@pleerock hello?

jotamorais:
> This could also be fixed by appending an empty handler to the list of middlewares when registering the action in the driver so that express stops handling the request.
> Also, is this project still maintained? This PR is 10 months old without any response from anyone but bots...
> @pleerock hello?

@Elassyo, I took over and I'm trying to first categorize and prioritize what can/need to be fixed first (that's why you see some bot activity lately). 
Ideas, comments, suggestions on how to do that are more than welcome - so are PRs :)

unix:
@jotamorais 
@Elassyo 
@pleerock 

Use bot to shut down prs & issues rather than deal with them, it's just a cover up, not a solution.
As a user & contributor, I feel very disappointed.

**I want every user to see these processes, and watch out for managers' inaction, and please use this opensource project very very carefully**

jotamorais:
> @jotamorais
> @Elassyo
> @pleerock
> 
> Use bot to shut down prs & issues rather than deal with them, it's just a cover up, not a solution.
> As a user & contributor, I feel very disappointed.
> 
> **I want every user to see these processes, and watch out for managers' inaction, and please use this opensource project very very carefully**

@unix, the idea to add the bot and to flag stale and close stale issues and PRs was to trigger action (and as you can see, triggered your action too).  We had over 150 open issues and 18 old PRs (some open since 2017).
I got permission to manage the Github and NPM's repositories late last October and I've already released 3 versions with important changes. 
Other changes with more potential to break the backward compatibility need to be more carefully evaluated and tested, understand if it makes sense to change, etc.

No one is trying to cover up or not deal with the issues (we didn't delete the issue - it's closed with a stale flag) and we will get back to it at some point. 
The idea is to use the bots to clear up the queue, bring the community together (not by trying to raise awareness with false allegations and "screaming").

I certainly appreciate your PR and any contribution, but there's no need to shout this kind of statements out in the wind. On the contrary, we should be calling for more ideas and contributions  on how to put the project back on track.

What are your suggestions?

PR #497, opened 2019-06-23
Include the current request context when resolving controller from IoC

adenhertog:
The `action` object is useful when resolving controller instances from the IoC provider, as the context is often used to add additional bindings for objects in a per-request lifecycle scope.

This change sends the `action` object as a second parameter in the `get()` method that's used when resolving controllers. This has been added as an optional parameter to achieve backwards compatibility.

The existing code also assumed that all IoC providers expose this `get()` method, which isn't always the case. An `IoCAdapter` interface has been added that can be implemented by consumers who use other providers.

Lastly the `class-validator` library has been split into a `peer-dependency` which closes https://github.com/typestack/routing-controllers/issues/384


NoNameProvided:
This issue has been automatically locked since there has not been any recent activity after it was closed. Please open a new issue for related bugs.

Issue #505, opened 2019-08-03
Anonymous middleware make it difficult to debug in express


saulotoledo:
Express allows debugging by setting `DEBUG=express:*` as described in [https://expressjs.com/en/guide/debugging.html](https://expressjs.com/en/guide/debugging.html).

After activating the debug messages, it is possible to see the middleware being called in the console as, for example:

```bash
express:router dispatching GET /v1/users/?page=1&limit=100 +3s
express:router query  : /v1/users/?page=1&limit=100 +2ms
express:router expressInit  : /v1/users/?page=1&limit=100 +2ms
express:router corsMiddleware  : /v1/users/?page=1&limit=100 +0ms
express:router <anonymous>  : /v1/users/?page=1&limit=100 +1ms
express:router <anonymous>  : /v1/users/?page=1&limit=100 +0ms
express:router <anonymous>  : /v1/users/?page=1&limit=100 +14ms
```

The previous example is from an application with routing-controllers and express. The anonymous functions are three middleware registered in the application by routing-controllers. Note that it is hard to know which middleware were triggered.

That happens because routing-controllers use anonymous functions as wrappers. We should be able to see the proper names in such logs.

saulotoledo:
@pleerock

GabeKuslansky:
You can achieve this functionality by registering middleware that logs this information. Example:

```
const app = createExpressServer({
  middlewares: [ResponseTimeLogger],
  controllers: [controllers]
});

@Middleware({
  type: "before"
})

export class ResponseTimeLogger implements ExpressMiddlewareInterface {
    use(req: Request, res: Response, next: NextFunction): void {
      const startTime = process.hrtime();
      res.on('finish', () => {
        const elapsedTime = process.hrtime(startTime);
        const elapsedTimeInMs = Math.round(elapsedTime[0] * 1000 + elapsedTime[1] / 1e6);
        console.log(`express:router ${req.method} ${req.path} +${elapsedTimeInMs}ms`);
      });
      next();
    };
  }
}
```

PR #506, opened 2019-08-03
Register named functions in express

saulotoledo:
Fixes https://github.com/typestack/routing-controllers/issues/505.

We are now able to see the names of the middlewares in the logs. This is the output for the same application used as an example in the reference issue.

```bash
express:router dispatching GET /v1/users/?page=1&limit=100 +10s
express:router query  : /v1/users/?page=1&limit=100 +1ms
express:router expressInit  : /v1/users/?page=1&limit=100 +2ms
express:router corsMiddleware  : /v1/users/?page=1&limit=100 +0ms
express:router HypermediaMiddleware  : /v1/users/?page=1&limit=100 +1ms
express:router BodyParser  : /v1/users/?page=1&limit=100 +0ms
express:router AuthorizationMiddleware  : /v1/users/?page=1&limit=100 +12ms
```

### DIFF
--- a/src/driver/express/ExpressDriver.ts
+++ b/src/driver/express/ExpressDriver.ts
@@ -53,18 +53,18 @@ export class ExpressDriver extends BaseDriver {
      * Registers middleware that run before controller actions.
      */
     registerMiddleware(middleware: MiddlewareMetadata): void {
+        let middlewareWrapper;
 
         // if its an error handler then register it with proper signature in express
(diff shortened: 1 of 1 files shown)


saulotoledo:
@jotamorais Could you take a look?

jotamorais:
> @jotamorais Could you take a look here?

Merged. Thanks for your contribution!

PR #512, opened 2019-09-28
updated dependencies

elthrasher:
The build/test/publish system for this repo is a bit outdated and many dependencies are out of date. It looks like the build and tests still work, but that gulpfile isn't working for me. I think the best bet is to just remove it and replace with npm scripts.

@jotamorais I was very happy to see that you were taking over maintenance of this project. Should I continue with this effort?


elthrasher:
Amazingly, I barely had to touch any code or tests. We now have 100% updated dependencies and `npm audit` comes up clean.

jotamorais:
@elthrasher Thanks for your PR. I don't think there will be any major problems either but I'm testing with a big project I'm currently working on (I think it will be a good "test ground").
I will get back to you shortly.

elthrasher:
@jotamorais I did some testing too and looks to me like breaking changes in `class-transformer` and `class-validator` might be an issue. Probably a good idea to move them to peer dependencies like has already been suggested: https://github.com/typestack/routing-controllers/pull/472 and https://github.com/typestack/routing-controllers/issues/384.
I can do that here if you'd like.

jotamorais:
> @jotamorais I did some testing too and looks to me like breaking changes in `class-transformer` and `class-validator` might be an issue. Probably a good idea to move them to peer dependencies like has already been suggested: #472 and #384.
> I can do that here if you'd like.

Yeah, let's do it. Then we can test the whole thing using peer dependencies and make sure it won't have any undesired effect.

PR #514, opened 2019-10-01
update changelog

elthrasher:
Didn't see anything else that should be called out, but here's a place to add stuff if need be.

Should we try publishing a preview version?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ### 0.8.0 [BREAKING CHANGES]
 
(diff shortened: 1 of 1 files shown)


jotamorais:
> Didn't see anything else that should be called out, but here's a place to add stuff if need be.
> 
> Should we try publishing a preview version?

Yeah, I was thinking of releasing this 0.8.0 as the preview version and something that we can use as a basis for all the other pending PRs and open issues. If we can publish the package and start using/testing, we will have a more solid ground to implement/fix other things as now we have more up-to-date depedencies and fix for some of the most critical issues I see the community complaining about.

PR #536, opened 2020-02-10
Add support for wildcard "all" routes

Elassyo:
I added support for routes which match all HTTP methods.

A change to the action type vs. request method check in `routeHandler` in the Express driver was required. However, being unfamiliar with the project source I'm not entirely sure this is the only location where such a change is required.

### DIFF
--- a/README.md
+++ b/README.md
@@ -733,7 +733,7 @@ There are set of prepared errors you can use:
 * UnauthorizedError
 
 
-You can also create and use your own errors by extending `HttpError` class.  
+You can also create and use your own errors by extending `HttpError` class.
 To define the data returned to the client, you could define a toJSON method in your error.
 
 ```typescript
@@ -755,7 +755,7 @@ class DbError extends HttpError {
         }
     }
 }
-``` 
+```
 
 #### Enable CORS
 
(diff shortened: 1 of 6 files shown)


jotamorais:
This PR has been merged onto **[next](https://github.com/typestack/routing-controllers/tree/next)** branch and a release candidate published - [0.9.0-alpha.1](https://github.com/typestack/routing-controllers/releases/tag/0.9.0-alpha.1).

@Elassyo, would you be so kind as to add some test coverage as well?

Thank you!

Elassyo:
Just added a test for the decorator, however I had to write a fix to prevent multiple routes from matching the same request, I put it in a separate PR (#568).
The test I added won't pass without it because the test route I added All(/users/me) also get matched by Get(/users/:id).

jotamorais:
> Just added a test for the decorator, however I had to write a fix to prevent multiple routes from matching the same request, I put it in a separate PR (#568).
> The test I added won't pass without it because the test route I added All(/users/me) also get matched by Get(/users/:id).

Thanks @Elassyo. I merged #568 in and this one as well in the [`next`](https://github.com/typestack/routing-controllers/tree/next) branch and published [0.9.0-alpha.2](https://github.com/typestack/routing-controllers/releases/tag/0.9.0-alpha.2) for tests before merging it into `master` and release final stable `0.9.0`.

Please, help with tests and let me know if you find something.

Thanks again!

PR #677, opened 2021-03-12
feat: add cast array to expected array query params

### LINKED ISSUE #123: Array parameters are not deserialized/validated properly

### LINKED ISSUE #320: array of  QueryParam problem
if I wan to parse a query param  to  string[] . for example:
```bash
GET /xxx?p=xx&p=yy
```
with router:  
```typescript
@Get('/xxx')
fun(@QueryParam('flags') flags: string[]){
}
```
will be ok, bug When 
```bash
GET /xxx?p=xx
```
will cause an error:
```
Given parameter p is invalid. Value (\"a\") cannot be parsed into JSON.
```

### LINKED ISSUE #389: QueryParams doesn't put single param values into an array
I'm currently on routingcontrollers v 0.7.6, using Typescript 2.7.2.  This issue may be better suited as a `class-transformer` bug, please advise if so.

## Initial Attempt: `QueryParam`
I originally had a controller that accepted an array of string values similar to [QueryParam with array](https://github.com/typestack/routing-controllers/issues/320).  When you specify more than one value in the query string, it works.  When you only specify one, it breaks.

Controller example:
```
    @Get('/route')
    public async myController(@QueryParam('value') values: string[]): Promise<ReturnModel> {
    }
```
working url example: `myurl/route?value=abc&value=def`
not working url example: `myurl/route?value=abc`

## Next Attempt: `QueryParams`
My next step was to implement the fix I found in issue 320, which involves defining a type and deconstructing all query params.

New Controller example:
```
    @Get('/route')
    public async myController(@QueryParams() { value }: SampleModel): Promise<ReturnModel> {
    }
```

SampleModel:
```
import { Type } from 'class-transformer';

export class SampleModel {
    @Type(( ) => String)
    public value: string[] = [];
}
```

## The Problem with arrays when using  `QueryParams`
This works, but this is also where my issue is.  If I pass in a multiple values, it works as expected.  If I pass in a single value, `value` becomes a string instead of a string array.  I'm trying to type this, since I'm using Typescript, so I tried changing the property type  in my model to:

`public value: string[] | string = [];`

and it works for the single parameter, but class transformer blows up with two or more of the same parameter name.

StackTrace (full paths redacted): 
```
newValue_1.push is not a function
application.js:630
    at ...\node_modules\class-transformer\TransformOperationExecutor.js:38:37
    at Array.forEach (<anonymous>)
    at TransformOperationExecutor.transform (...\node_modules\class-transformer\TransformOperationExecutor.js:30:20)
    at _loop_1 (...\node_modules\class-transformer\TransformOperationExecutor.js:154:46)
    at TransformOperationExecutor.transform (...\node_modules\class-transformer\TransformOperationExecutor.js:178:18)
    at ClassTransformer.plainToClass (...\node_modules\class-transformer\ClassTransformer.js:17:26)
    at Object.plainToClass (...\src\index.ts:37:29)
    at ActionParameterHandler.transformValue (...\node_modules\routing-controllers\ActionParameterHandler.js:142:42)
    at ActionParameterHandler.normalizeParamValue (...\node_modules\routing-controllers\ActionParameterHandler.js:113:35)
    at ActionParameterHandler.handle (...\node_modules\routing-controllers\ActionParameterHandler.js:36:27)
```

Preferred Results, in order of preference:
1) Allow what the OP tried in Issue 320-- this is the way it works in some other model binding systems, such as ASP.NET.
2) If I'm using my second approach posted here, and I there's only one query parameter for something that should be an array (my `SampleModel.value` property), place that single value into an array.
3) Allow for multi-types so I can correctly type my model's properties in my apps.

## Workaround
The workaround isn't horrible, but it would be nice to have the underlying frameworks updated so we don't have to repeat the workarounds in numerous clients.

My workaround:
```
    // if only one value is passed in, the type will be a string. convert to an array for consistency in handling.
        const values: string[] = Array.isArray(value) ? value: [value] || [ ];
```

BenjD90:
## Checklist

  - `npm run lint:check` passes
- [x] I have run the project locally and verified that there are no errors → tests pass

### Fixes
Fixes #389, Fixes #320, Fixes #123

The idea is to parse query parameters as it is used to be. I added some exemple in the README, and in the tests :

You can use `isArray` option to get a query param array. This will cast the query param :

```typescript
@Get("/users/by-multiple-ids")
getUsers(@QueryParam("ids", { isArray: true}) ids: string[]) {
}
```

`GET /users/by-multiple-ids?ids=a` → `ids = ['a']`
`GET /users/by-multiple-ids?ids=a&ids=b` → `ids = ['a', 'b']`

You can combine use `isArray` option with `type` option to get a query param array of one type. This will cast the query param :

```typescript
@Get("/users/by-multiple-ids")
getUsers(@QueryParam("ids", { isArray: true, type: Number}) ids: number[]) {
}
```

`GET /users/by-multiple-ids?ids=1` → `ids = [1]`
`GET /users/by-multiple-ids?ids=1&ids=3.5` → `ids = [1, 3.5]`

For QueryParams : 

```typescript
enum Roles {
    Admin = "admin",
    User = "user",
    Guest = "guest",
}

class GetUsersQuery {

    @IsPositive()
    limit: number;

    @IsAlpha()
    city: string;

    @IsEnum(Roles)
    role: Roles;

    @IsBoolean()
    isActive: boolean;

    @IsArray()
    @IsNumber(undefined, { each: true })
    @Type(() => Number)
    ids: number[];
}

@Get("/users")
getUsers(@QueryParams() query: GetUsersQuery) {
    // here you can access query.role, query.limit
    // and others valid query parameters
    // query.ids will be an array, of numbers, even with one element
}

### DIFF
--- a/README.md
+++ b/README.md
@@ -378,6 +378,28 @@ getUsers(@QueryParam("limit") limit: number) {
 }
 ```
 
+You can use `isArray` option to get a query param array. This will cast the query param :
+
+```typescript
+@Get("/users/by-multiple-ids")
+getUsers(@QueryParam("ids", { isArray: true}) ids: string[]) {
+}
+```
+
+`GET /users/by-multiple-ids?ids=a` → `ids = ['a']`
+`GET /users/by-multiple-ids?ids=a&ids=b` → `ids = ['a', 'b']`
+
+You can combine use `isArray` option with `type` option to get a query param array of one type. This will cast the query param :
+
+```typescript
+@Get("/users/by-multiple-ids")
+getUsers(@QueryParam("ids", { isArray: true, type: Number}) ids: number[]) {
+}
+```
+
+`GET /users/by-multiple-ids?ids=1` → `ids = [1]`
+`GET /users/by-multiple-ids?ids=1&ids=3.5` → `ids = [1, 3.5]`
+
 If you want to inject all query parameters use `@QueryParams()` decorator.
 The bigest benefit of this approach is that you can perform validation of the params.
 
@@ -402,12 +424,17 @@ class GetUsersQuery {
     @IsBoolean()
     isActive: boolean;
 
+    @IsArray()
+    @IsNumber(undefined, { each: true })
+    @Type(() => Number)
+    ids: number[];
 }
 
 @Get("/users")
 getUsers(@QueryParams() query: GetUsersQuery) {
     // here you can access query.role, query.limit
     // and others valid query parameters
+    // query.ids will be an array, of numbers, even with one element
 }
 ```
 
@@ -1535,7 +1562,7 @@ export class QuestionController {
 
 | Signature                                                        | Example                                           | Description                                                                                                                                 |
 | ---------------------------------------------------------------- | ------------------------------------------------- | ------------------------------------------------------------------------------------------------------------------------------------------- |
-| `@Authorized(roles?: string\|string[])`                          | `@Authorized("SUPER_ADMIN")` get()                | Checks if user is authorized and has given roles on a given route. `authorizationChecker` should be defined in routing-controllers options. |  |
+| `@Authorized(roles?: string\|string[])`                          | `@Authorized("SUPER_ADMIN")` get()                | Checks if user is authorized and has given roles on a given route. `authorizationChecker` should be defined in routing-controllers options. |
 | `@CurrentUser(options?: { required?: boolean })`                 | get(@CurrentUser({ required: true }) user: User)  | Injects currently authorized user. `currentUserChecker` should be defined in routing-controllers options.                                   |
 | `@Header(headerName: string, headerValue: string)`               | `@Header("Cache-Control", "private")` get()       | Allows to explicitly set any HTTP header returned in the response.                                                                          |
 | `@ContentType(contentType: string)`                              | `@ContentType("text/csv")` get()                  | Allows to explicitly set HTTP Content-Type returned in the response.                                                                        |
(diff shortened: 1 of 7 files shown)


BenjD90:
I don't understand the semantic release check fail :confused: If anyone has an idea, I'll take it !

BenjD90:
Fixes #389, #320, #123

BenjD90:
@NoNameProvided @nolazybits Is it possible to review this PR ?

thanks,
Benjamin

nolazybits (approved):
Looks good to me but that would be my first review here. I might have missed something. @NoNameProvided might want to double check + I don't think I have write access to this repo (required to have a vlad approval) to it won't help unblock this MR

BenjD90:
I found a bug on boolean management if only one is sent as query param, I'll complete the test with more boolean cases

BenjD90:
The bug is now fixed :smiley: 
I have added the test `@QueryParams should give a proper values from request's query with validate whitelist option on` for this specific case.

jotamorais (approved):

PR #702, opened 2021-04-08
fix(deps): fix lockfiles

loriswit:
## Description
I was trying to fix some stuff in this repo, but I was having trouble installing the dependencies.

First, there are many weird URLs in the lock files that try to download dependencies from `npm.internal.equipmentshare.com`, which I can't personally access, so I changed them to the conventional `registry.npmjs.org`.

Also, it looks like *yarn.lock* is **not synced** with *package-lock.json* (for example, *package-lock.json* has @babel/core version 7.11.1, but *yarn.lock* has version 7.12.9). Since *yarn.lock* has only been recently added to the repository (December 2020), I supposed it was generated by mistake, so I just deleted it (I don't think we need both *package-lock.json* **and** *yarn.lock* anyway, right?).

## Checklist

  - `npm run lint:check` passes


jotamorais (approved):
# @

PR #739, opened 2021-07-01
fix: Check if a number is NaN in ActionParameterHandler

### LINKED ISSUE #348: @Param default validation
Hi,

I found that `This technique works not only with @Body but also with @Param, @QueryParam, @BodyParam and other decorators.` in this validation [docs article](https://github.com/typestack/routing-controllers#auto-validating-action-params).

But by default `@Param('id') id: number` is not validated and accepts `NaN` as value in case of invalid `id`.
There is no validate option for `@Param` decorator, which mentioned in the docs.

So mb you need a fix to be with your library docs.

Thanks.

borjapazr:
## Description

`@Param('id') id: number` is not validated and accepts `NaN` as value in case of invalid `id`.

For example, if the following URL is accessed (`http://example.com/users/asdf`), instead of throwing an error, the `id` value is parsed to `NaN`. `InvalidParamError(value, parameterName, parameterType)` should be thrown instead.

The problem is due to how the checking of the result of the cast operation is done. `NaN` is not equal to `NaN`. Because of this it is necessary to use `Number.isNaN()`.

**Example**

```
@JsonController('/users')
class UserController {

  @Get('/:id')
  get (@Param('id') id: number) {
    console.log(id);
  }
}
```

**Result**

`GET http://example.com/users/1234` -> `1234`
`GET http://example.com/users/asdf` -> `NaN`

## Checklist

  - `npm run lint:check` passes

### Fixes

fixes #348


borjapazr:
Hi! Could you review this PR, please? 😊

Lxxyx (approved):

borjapazr:
Another updated PR has been created #815

PR #743, opened 2021-07-04
fix: Fixed example in README.md

### LINKED ISSUE #742: fix: Fixed example in README.md
### Description

(not sure if bug-report is appropriate for a bug in documentation)

I was trying to copy-paste an example from README.md, and it resulted in compilation errors

**Minimal code-snippet showcasing the problem**
 

```ts
export const fileUploadOptions = () => {
    storage: multer.diskStorage({
        destination: (req: any, file: any, cb: any) => { ...
        },
        filename: (req: any, file: any, cb: any) => { ... // Compilation error here
        }
    }),
    fileFilter: (req: any, file: any, cb: any) => { ...
    },
    limits: {
        fieldNameSize: 255,
        fileSize: 1024 * 1024 * 2 // Compilation error here
    }
};
```

### Expected behavior
There shouldn't be a compilation when copy-pasting the example

### Actual behavior
There is a compilation error. This is because JS/TS parser thinks that this is a normal lambda ( eg `() => { ... console.log and etc ... }` ), with labels inserted in weird places.
Most likely the intent was to return an object instead. To do so, this "code-block" must be incapsulated in (), eg `() => ({ object: true })`.

ArtemLavrentii:
Most likely the intent of this lambda was to return an object, and not to have a code-block with labels.

## Checklist

- [ ] the code follows the established code style of the repository
  - `npm run lint:check` passes
- [ ] I have run the project locally and verified that there are no errors

### Fixes

fixes #742

### DIFF
--- a/README.md
+++ b/README.md
@@ -507,7 +507,7 @@ You can also specify uploading options to multer this way:
 
 ```typescript
 // to keep code clean better to extract this function into separate file
-export const fileUploadOptions = () => {
+export const fileUploadOptions = () => ({
     storage: multer.diskStorage({
         destination: (req: any, file: any, cb: any) => { ...
         },
@@ -520,7 +520,7 @@ export const fileUploadOptions = () => {
         fieldNameSize: 255,
         fileSize: 1024 * 1024 * 2
     }
-};
+});
 
 // use options this way:
 @Post("/files")


ArtemLavrentii:
I'm not 100% that even this is a correct example. Shouldn't upload options just be an object ?

jotamorais (approved):

PR #780, opened 2021-09-30
Fix: koa-views always return the result of the first execution

nanyuantingfeng:
After the first rendering of html, all request results return the first rendered html;

The reason is that `this.koa.use (FN)`  is called many times, resulting in subsequent requests going through the first registration `FN`; The closure variables(renderOptions) ref to the first FN`s renderOptions; Results in the return of the same html.


beclass:
@nanyuantingfeng I don't think so now. How to solve it? Modify the source code?

PR #845, opened 2022-02-17
fix: check if headersSent

Leechael:
## Description

Check `response.headersSent` in both handleSuccess & handleError, prevents that response already sent and Express app crashed.

### DIFF
--- a/src/driver/express/ExpressDriver.ts
+++ b/src/driver/express/ExpressDriver.ts
@@ -252,7 +252,7 @@ export class ExpressDriver extends BaseDriver {
    * Handles result of successfully executed controller action.
    */
   handleSuccess(result: any, action: ActionMetadata, options: Action): void {
-    if (action.response && action.response.headersSent) {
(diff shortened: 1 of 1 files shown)


attilaorosz:
While I agree with this, I feel like we should somehow indicate that your response from the controller will not be delivered. Simply refusing to return your response from the controller might cause a lot of confusion and debugging. @NoNameProvided thoughts?

attilaorosz:
I get your point and I agree with it, my only concern is that it might cause unexpected results. We should somehow indicate (log?) that this happened. For example just adding and error.log letting the dev know that the controller result is ignored because the headers are already sent. Obviously console.log is not the perfect solution, just a low hanging fruit :)

Leechael:
I'm agree to improve that, how about optional warning to console.log? and the warning and opt-out.

attilaorosz:
Could you please add a tests?

Leechael:
> Could you please add a tests?

Sure. Let me check how to do that.

PR #860, opened 2022-03-21
Passing options with UseInterceptor

driescroons:
## Description
When using routing controllers, I don't want to explicitly set an interceptor to each controller action even though I (might) want to specify the priority, in how they're applied to the controller action.

For example; I have a:
- Flush interceptor (for mikro-orm)
- An Interceptor that formats my controller action response (using class-validator and class-transformer)

When I now specify the flush interceptor on global or controller level, it only get's executed after my response has been formatted. I want the flush to happen before any controller action response formatting happens, without having to specify the flush interceptor, on every controller action, below the response formatting interceptor. 

This PR suggests a way to pass and specify options (in this case mostly for the priority prop) to the interceptors, and sorts them according to importance before being applied to each controller action.

This PR still requires tests and additional docs, but I want to get some opinions whether this is something that could be merged. 

Local development setup used: https://github.com/typestack/routing-controllers/issues/861
Project I tested this with (above setup required): https://github.com/driescroons/routing-controllers-interceptor-priority-testing

## Checklist

- [X] the pull request title describes what this PR does (not a vague title like `Update index.md`)
- [X] the pull request targets the *default* branch of the repository (`develop`)
- [X] the code follows the established code style of the repository
  - `npm run lint:check` passes
- [X] I have run the project locally and verified that there are no errors

### DIFF
--- a/src/RoutingControllers.ts
+++ b/src/RoutingControllers.ts
@@ -172,6 +172,7 @@ export class RoutingControllers<T extends BaseDriver> {
    * Creates interceptors from the given "use interceptors".
    */
   protected prepareInterceptors(uses: InterceptorMetadata[]): Function[] {
+    uses.sort((interceptor1, interceptor2) => interceptor2.priority - interceptor1.priority);
     return uses.map(use => {
       if (use.interceptor.prototype && use.interceptor.prototype.intercept) {
         // if this is function instance of InterceptorInterface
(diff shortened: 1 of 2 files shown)


driescroons:
Glad that you bring that up @attilaorosz as I've already been working on that PR as well! A good reason why I think we also need this for middleware is when we want to run authentication before any validation (which is not the case right now), or other middleware, without having to specify it in a specific order on the controller action.

 I'll add some tests and docs to this PR!

PR #1114, opened 2023-04-19
Fix: Remove trailing slash from route path in KoaDriver

### LINKED ISSUE #1113: Breaking Change in 0.10.0, @koa/router and trailing slash
### Description
`routing-controllers` registers all routes with trailing slashes (eg. `/user/`). Updated router do not automagically removes the trailing slash. Meaning accessing routes without trailing slash return 404.

Related issue: https://github.com/koajs/router/issues/79

### Expected behavior
```
GET /user -> 200
GET /user/ -> 200 (as previously, or 404)
```

### Actual behavior
```
GET /user -> 404
GET /user/ -> 200
```

Aareksio:
## Description
Before the update, `@koa/router` allowed path declared with a trailing slash to be accessible both with or without trailing slash
```ts
router.get('/user/', handler)

// GET /user -> 200
// GET /user/ -> 200
```

After the update, only trailing slash path is accessible. This makes registering nested routes with `@Get('/')` infeasible, when one does not use trailing slashes for URLs.

```ts
@JsonController('/user')
export class UserController {
  @Get('/') // <-- this causes the route to become `/user/`
  getUser() {}
}
```

Proposed change, restricted in scope to affected KoaDriver only, removes the trailing slash if one appears in the route (aside from `/`). `@koa/router` allows such paths to be accessed both with and without trailing slash, as before.

Please note `@Get('/')` is what official tests are using, and it works just fine with express:
https://github.com/typestack/routing-controllers/blob/develop/test/functional/controller-base-routes.spec.ts#L18-L21

## Checklist

  - `npm run lint:check` passes

### Fixes

Fixes #1113

### DIFF
--- a/src/driver/koa/KoaDriver.ts
+++ b/src/driver/koa/KoaDriver.ts
@@ -129,7 +129,12 @@ export class KoaDriver extends BaseDriver {
     const afterMiddlewares = this.prepareMiddlewares(uses.filter(use => use.afterAction));
 
     // prepare route and route handler function
-    const route = ActionMetadata.appendBaseRoute(this.routePrefix, actionMetadata.fullRoute);
+    let route = ActionMetadata.appendBaseRoute(this.routePrefix, actionMetadata.fullRoute);
+
+    if (route.length > 1 && route.endsWith('/')) {
+      route = route.substring(0, route.length - 1)
+    }
+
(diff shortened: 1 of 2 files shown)


attilaorosz:
This is a nice catch, thanks for the contribution!

attilaorosz (approved):

PR #1375, opened 2024-04-15
feat: enhanced typing for CurrentUserChecker by introducing generics

### LINKED ISSUE #1374: feature: Introduce Generics to CurrentUserChecker for Enhanced Type Safety
### Description
Currently, the `CurrentUserChecker` function in `routing-controllers` uses a return type of `Promise<any> | any`, which doesn't utilize TypeScript's capabilities for strong typing. This can lead to potential type mismatches and less predictable code, especially in larger projects where strict type checking is crucial.

### Proposed Solution
The `CurrentUserChecker` should be enhanced to support generic types, allowing developers to specify the expected user type more explicitly. This will improve code reliability and developer experience by leveraging TypeScript's type system more effectively.

#### Implementation:
- Modify the type definition of `CurrentUserChecker` to: `export type CurrentUserChecker = <User = any>(action: Action) => Promise<User | null> | User | null;`
- This change will allow the function to return `User | null` or `Promise<User | null>`, where `User` is a generic type that can be defined by the developer.

#### Alternative Solutions Considered:
- An alternative would be to enforce a stricter non-nullable type by default, but this could introduce breaking changes for existing users who may rely on the flexibility of returning `any`.
- Another option could be to provide optional configuration settings in the library to enable strict typing, but this would complicate the library's configuration and usage.

The proposed solution strikes a balance between improving type safety and maintaining backward compatibility, thus providing a seamless upgrade path for current users while offering improved functionality.

angelxmoreno:
## Overview
This PR introduces a change to the `CurrentUserChecker` type definition to support generics, allowing for more precise type inference and usage in projects utilizing `routing-controllers`.

## Changes
- The `CurrentUserChecker` type definition has been modified to accept a generic type parameter `User`. This parameter defaults to `any`, ensuring backward compatibility.
- The return type has been updated to include `User | null` in addition to `Promise<User | null>`. This change enhances type safety by allowing the function to explicitly specify a nullable user object, which aligns better with TypeScript practices where explicit types are preferable for nullability.

## Motivation
Previously, `CurrentUserChecker` was typed to return `any` or `Promise<any>`, which did not leverage TypeScript's full capabilities for type safety and led to less precise type inference in consuming code. By introducing generics, developers can now specify the expected user object type, leading to cleaner and more maintainable codebases.

## Compatibility
The default generic type is `any`, which maintains the current behavior for existing codebases using `routing-controllers`. This update is fully backwards compatible and does not require any changes to existing implementations unless the consuming codebase opts into using generics for more specific typing.

## Example
With the updated `CurrentUserChecker`, developers can now define the expected return type explicitly:

```typescript
const currentUserChecker: CurrentUserChecker<MyUserType> = (action: Action) => {
  // Implementation that returns MyUserType or null
};
```
This ensures that the type of the user object is known and checked at compile time, reducing runtime errors and improving the developer experience.

## Checklist
- [x] `npm run prettier:check` passes
- [x] `npm run lint:check` passes

### Fixes
fixes #1374

### DIFF
--- a/src/CurrentUserChecker.ts
+++ b/src/CurrentUserChecker.ts
@@ -3,4 +3,4 @@ import { Action } from './Action';
 /**
  * Special function used to get currently authorized user.
  */
-export type CurrentUserChecker = (action: Action) => Promise<any> | any;
+export type CurrentUserChecker = <User = any>(action: Action) => Promise<User | null> | User | null;
--- a/src/index.ts
+++ b/src/index.ts
@@ -89,6 +89,8 @@ export * from './CustomParameterDecorator';
 export * from './RoleChecker';
 export * from './Action';
 export * from './InterceptorInterface';
+export * from './CurrentUserChecker';
+export * from './AuthorizationChecker';
 
 export * from './driver/BaseDriver';
 export * from './driver/express/ExpressDriver';


attilaorosz (approved):

attilaorosz:
Thanks for the contribution!

PR #1482, opened 2025-01-21
fix: Render decorator for Koa Driver

### LINKED ISSUE #602: Help with Render()
**I was trying to...**
I am trying to use Render() with EJS.

**The problem:**
First off I want to thank you for this library, it really awesome and have really enjoyed using it.

I have a new project I am working on and decided that EJS would be best for it. I am using the Render decorator and the first call it works perfectly but any subsequent calls that use a different template will get the first template returned.

For example, if I request `/puppies`  it will return the template from puppies.ejs. If I then request `/kittens`, debug shows the template file for puppies still being returned.

I am using the koa-ejs module with this, but have tested with koa-views and get the same result.

Here are my files:
```typescript
// server.ts
import { createKoaServer } from 'routing-controllers';
import Koa from 'koa';
import * as HttpStatus from 'http-status-codes';
import render from 'koa-ejs';
import * as path from 'path';
import 'reflect-metadata';

const app: Koa = createKoaServer({
  controllers: [__dirname + "/controllers/**/*{.ts,.js}"]
});

render(app, {
  root: path.join(__dirname, 'views'),
  layout: 'layout',
  viewExt: 'ejs',
  cache: false,
  debug: true
});

app.listen(3001), () => {
  console.log(`Server listening on port 3001`)
};



// kittens.controller.ts
import {Controller, Get, Render } from 'routing-controllers';

@Controller('/kittens')
export class KittensController {
    constructor(){}

    @Render('kittens')
    @Get()
    getKittens() {
        return {
            animal: 'kittens'
        }
    }
}



//puppies.controller.ts
import {Controller, Get, Render} from 'routing-controllers';

@Controller('/puppies')
export class PuppiesController {
    constructor(){}

    @Get()
    @Render('puppies')
    getPuppies() {
        return {
            animal: 'puppies'
        }
    }
}
```

Templates:

```
<!-- layout.ejs -->
<!DOCTYPE html>
<html lang="en">
<head>
  <meta charset="UTF-8">
  <title>EJS Test</title>
</head>
<body>

  <%- body %>

</body>
</html>



<!-- puppies.ejs -->
This is the puppies  page.


<!-- kittens.ejs -->
This is the kittens  page.

```

I am a network engineer by day, learning app development by night so if my code doesn't look good feel free to point it out.

Thank you very much for your time and any help you may be able to provide.

### LINKED ISSUE #610: 【BUG】koa render will break all API request
When use `@Render` in Koa, it will break all api request, all API request will return a html template.

## How to produce
1. Create a html render controller
```
// html/default.ts
@Controller()
export default class DefaultController {
  @Get('/homepage')
  @Render('home/index')
  async index() {
    return {
      title: config.app.name,
    };
  }
}
```
2. Create a api json controller
```
// api/default.ts
@JsonController()
export default class DefaultController {
  @Get("/api/users")
  getAll() {
    return {
      code: 0,
      data: [{ name: 'Jim Green', id: 1 }]
    };
  }
```

3. Test
```
curl http://localhost/api/users    // show json as expected
curl http://localhost/homepage   // show HTML page as expected
curl http://localhost/api/users    // show HTML page (!!NOT EXPECTED!!)
```

## Reason code
https://github.com/typestack/routing-controllers/blob/24556885e58e1022531881eecb963c363e9933a9/src/driver/koa/KoaDriver.ts#L245-L251

in KoaDriver.ts, handleSuccess will register a render middleware. so that every request will goes to the middleware, even a JSON api.

### LINKED ISSUE #902: fix: empty page on @Render() responds with 404
### Description

When the page specified in the `@Render()` decorator returns an empty string, the server responds with the status code 404 (Not Found).

**Minimal code-snippet showcasing the problem**
 

```ts
import { Controller, Get, Render } from 'routing-controllers';

@Controller()
export class MyController {
    @Get()
	@Render('index.pug') // index.pug must be empty
	index() {
		
	}
}
```

### Expected behavior

The server should send an empty body and respond with 200 (OK).

### Actual behavior

The server sends the error page and responds with 404 (Not Found).

### LINKED ISSUE #1147: fix: Wrong logic for render decorator
### Description
I added a `@Render()` decorator for one of my routes， but other routes were also affected. I debuged into the code, and in `KoaDriver`, I found this:

```typescript

    else if (action.renderedTemplate) {
      // if template is set then render it // TODO: not working in koa
      const renderOptions = result && result instanceof Object ? result : {};

      this.koa.use(async function (ctx: any, next: any) {
        await ctx.render(action.renderedTemplate, renderOptions);
      });
    }
```

This could result in the following two results: 
1. Affect other routes which are `not` marked  with `@Render` decorator.
2. Register a middleware for koa for every `@Render` request.

### Solution
I suggest change this code to this.

```typescript
       else if (action.renderedTemplate) {
            const renderOptions = result && result instanceof Object ? result : {};
            return Promise.resolve(
                    options.context.render(action.renderedTemplate, renderOptions)
                ).then(() => options.next())
        }
```

Telokis:
## Description
Fixes the `@Render` decorator when using `Koa` by hijacking the `next` function to ensure the `ctx.render` promise will be completed.

## Checklist

  - `npm run lint:check` passes

### Fixes
fixes #1147, fixes #902, fixes #610, fixes #602


attilaorosz:
Unfortunately I’m not familiar with koa so I cannot review this.
@NoNameProvided any suggestions?

Telokis:
> Unfortunately I’m not familiar with koa so I cannot review this. @NoNameProvided any suggestions?

@attilaorosz Hey, thanks for replying!

I can give more details:
- First of all, the previous code (`this.koa.use`) was very wrong and should absolutely be removed even if it's not replaced. It registers a new middleware that will always call `ctx.render`, it's not for the current route or handling but forever. As soon as one `@Render` route is called, the middleware is added and subsequently used.
- The `next` function of Koa is a Promise. Our `ctx.render` also is a promise that must be `await`-ed but the current function is not `async`. In order not to alter the flow too much, I use a trick to redefine the `next` function (that will eventually be `await`-ed) so it runs the `ctx.render` first.

I've already implemented this solution on my local version to confirm it works and the tests are also there to provide some insurance about it.

Telokis:
@attilaorosz @NoNameProvided Hello!

Can I get some feedback on this? It's a bit annoying to manually edit the file everytime I reinstall my modules so the feature can work as expected! haha

Thanks!

attilaorosz:
Sorry for the late response!
While I generally agree with fixing this, my main concern is that while the old koa-views seemed to support any kind of templating engine, koa-ejs only supports ejs. I'm not sure if anyone is using anything. Might be a good idea to provide a bridge between routing-controllers and whatever templating engine instead.

Telokis:
`koa-views` can still be used. (The [new one](https://www.npmjs.com/package/@ladjs/koa-views) because the [old one](https://www.npmjs.com/package/koa-views) is deprecated)  
`routing-controller` didn't make any choice regarding this, anyway. With or without my PR, no engine is present or configured by default.  

Without my PR, though, using `@Render` at all will break the app as soon as the decorated controller is visited.

attilaorosz:
Sounds good. Sorry, as I mentioned I'm not experience with koa. I'm open to do this but then let's remove the koa-views dev dependency.

Telokis:
> Sounds good. Sorry, as I mentioned I'm not experienced with koa. I'm open to do this but then let's remove the koa-views dev dependency.

@attilaorosz No worries, it's fine! I'll remove the dev dependency, indeed I missed it!

attilaorosz:
Thanks for the contribution!

Telokis:
> Thanks for the contribution!

@attilaorosz No problem, thanks for merging! Do you think it will be released soon?

attilaorosz:
Releasing right now

PR #1483, opened 2025-01-29
Update package.json to change the cookie package vulnerable version 

yougharta:
I just updated the cookie package version from 0.6.0 to 0.7.2  , the version you are using is vulnerable to this : https://nvd.nist.gov/vuln/detail/CVE-2024-47764

Thank you

### DIFF
--- a/package.json
+++ b/package.json
@@ -39,7 +39,7 @@
     "test:debug": "NODE_ENV= node --inspect=0.0.0.0:20001 ./node_modules/.bin/jest --colors --runInBand --no-coverage --watch"
   },
   "dependencies": {
-    "cookie": "^0.6.0",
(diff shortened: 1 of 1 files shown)


attilaorosz:
Thanks for the contribution. Unfortunately it seems like the package-lock file was nor included in the change

yougharta:
hello @attilaorosz , thanks for replying
if i include it (package-lock.json) and send the pull request again , does it going to work ?

attilaorosz:
You can just add another commit with the package lock. Without that the build fails.

yougharta:
hello @attilaorosz  , can you check this [pull](https://github.com/typestack/routing-controllers/pull/1484) request now

